PR #18, opened 2023-06-16
Utils

heeckhau:


heeckhau:
@sinui0 is this what you had in mind?

I initially renamed the `utils` folders to `mpz_utils`, but reverted that to make the change smaller. I am not sure what the best choice is here.

sinui0:
I think in the near term we can stick with the `tlsn-utils` repo. When you separated it out of the `tlsn` repo that took care of the biggest problem. Now that we use `itybity` I think we can get rid of the `tlsn-utils` dep, and we just need to decide what we want to do with `tlsn-utils-aio`

heeckhau:
@sinui0 I squashed and rebased the branch onto the latest `dev` branch.

This indeed removed a lot of uses of `tlsn-utils`.

I could not entirely remove it, because there a few (5) uses left of: `utils::id::NestedId`, `utils::iter::xor` and `utils::iter::FilterDrain`

Good enough to merge?
If we do, we also need to update the `tlsn-utils` repo. This branch currently uses the [cleanup](https://github.com/tlsnotary/tlsn-utils/tree/cleanup) branch.

sinui0:
I'm not sure we want to go ahead with this, as we may need the `non_blocking_backend` stuff for `tlsn` as well. For now I think we should just keep `mpz` depending on those `tlsn-utils*` crates until we properly modularize them and possibly release.

PR #27, opened 2023-07-05
Matrix transpose bitorder

sinui0:
This PR flips the bit order of the resulting transposed matrix

### DIFF
--- a/matrix-transpose/src/lib.rs
+++ b/matrix-transpose/src/lib.rs
@@ -119,6 +119,37 @@ mod tests {
     }
 
     #[test]
+    fn test_transpose_naive() {
+        let matrix = [
+            // ------- bits in lsb0
+            3u8,   // 1 1 0 0 0 0 0 0
+            76u8,  // 0 0 1 1 0 0 1 0
+            120u8, // 0 0 0 1 1 1 1 0
+            9u8,   // 1 0 0 1 0 0 0 0
+            17u8,  // 1 0 0 0 1 0 0 0
+            102u8, // 0 1 1 0 0 1 1 0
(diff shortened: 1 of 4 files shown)


th4s (approved):
ack :+1:

matrix-transpose/src/lib.rs
```diff
@@ -87,17 +88,34 @@ mod tests {
         (0..elements).map(|_| rng.gen::<T>()).collect()
     }
 
+    fn transpose_naive(data: &[u8], row_width: usize) -> Vec<u8> {
+        use itybity::*;
+
+        let bits: Vec<Vec<bool>> = data.chunks(row_width).map(|x| x.to_lsb0_vec()).collect();
```
themighty1, line 94: isn't `row_width` the width in **bits** ? here it is treated as if it was in **bytes**

themighty1 (changes requested):

matrix-transpose/src/lib.rs
th4s: I think it is correct here to use the row width in bytes, because we iterate in chunks of byte rows and turn this chunk into bit rows. So in the beginning it is bytes and after this operation it is bits.

themighty1 (changes requested):
some tests seem to be failing

Is there an easy way to test the "simd-transpose" feature from command line? I couldnt find one. 
Maybe we should have our CI to also run the tests with  "simd-transpose" enabled?

sinui0:
I don't see any failing tests, try `cargo clean` then `cargo +nightly t --features simd-transpose`

sinui0:
I'm going to merge this, it is blocking #37. Changes to CI etc can be done in another PR

PR #28, opened 2023-07-05
block refactor, fixed key aes, cointoss

sinui0:
This PR refactors some of our core types, and adds a simple cointoss protocol.

Mostly its a few odds and ends from the in-progress OT work.

# Changes
1. Replace the internal type of `Block` to be a `[u8; 16]`. `u128` was a mistake, blocks shouldn't be interpreted as integers but as a bag of bits. Eventually we will want to replace this with a type providing vectorized operations where possible.
2. Introduces cointoss protocol, needed for KOS
3. Pulls out some of the methods on `Block` into `FixedKeyAes`.


mpz-core/src/lib.rs
```diff
@@ -2,11 +2,19 @@
...
 pub mod utils;
 pub mod value;
 
 pub use block::{Block, BlockSerialize};
+
+/// A protocol with a message type.
+pub trait ProtocolMessage {
+    /// The type of message used in the protocol.
+    type Msg: Send + Sync + std::fmt::Debug + 'static;
+}
```
th4s, line 20: Do you intend to add methods to this trait at some later point?

mpz-core/src/cointoss.rs
th4s, line 1: Can we split that into `sender` and `receiver` modules?

mpz-core/src/cointoss.rs
```diff
@@ -0,0 +1,293 @@
...
+                actual: receiver_seeds.len(),
+            });
+        }
+
+        if commitments.len() != receiver_seeds.len() {
+            return Err(CointossError::CountMismatch {
+                expected: commitments.len(),
+                actual: receiver_seeds.len(),
+            });
+        }
```
th4s: Isn't this already checked in `reveal`?

th4s (approved):
Nice :rocket:

themighty1:
I get compilation errors when running `cargo test`

mpz-core/src/lib.rs
sinui0: No, this is simply to associate a message type with something implementing a protocol

sinui0:
> I get compilation errors when running `cargo test`

This is expected, as this PR is scoped only to `mpz-core`. Following PRs will refactor dependents to accomodate these changes. You can test `mpz-core` in isolaton

sinui0:
I've slightly modified the cointoss to commit to all seeds at once, this is more efficient.

mpz-core/src/aes.rs
```diff
@@ -0,0 +1,45 @@
+//! Fixed-key AES cipher
+
+use aes::{cipher::generic_array::GenericArray, Aes128};
+use cipher::{generic_array::functional::FunctionalSequence, BlockEncrypt, KeyInit};
+use once_cell::sync::Lazy;
+
+use crate::Block;
+
+/// AES fixed key
```
themighty1: /// A fixed AES key (arbitrarily chosen)

mpz-core/src/aes.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Fixed-key AES cipher
+pub struct FixedKeyAes {
+    aes: Aes128,
+}
+
+impl FixedKeyAes {
+    /// Tweakable circular correlation-robust hash function instantiated
+    /// using fixed-key AES.
+    ///
+    /// See <https://eprint.iacr.org/2019/074>
```
themighty1: See <https://eprint.iacr.org/2019/074> (section 7.4)

mpz-core/src/aes.rs
```diff
@@ -0,0 +1,45 @@
...
+    ///
+    /// `π(π(x) ⊕ i) ⊕ π(x)`
```
themighty1: /// `π(π(x) ⊕ i) ⊕ π(x)` , where π can be instantiated using AES with a fixed key.

mpz-core/src/block.rs
```diff
@@ -1,35 +1,35 @@
...
 use core::ops::{BitAnd, BitXor};
-use itybity::{BitLength, GetBit, Lsb0, Msb0};
-use rand::{CryptoRng, Rng};
+use itybity::{BitIterable, BitLength, GetBit, Lsb0, Msb0};
+use rand::{distributions::Standard, prelude::Distribution, CryptoRng, Rng};
 use serde::{Deserialize, Serialize};
-use std::convert::{From, TryInto};
+use std::convert::From;
 
 /// A block of 128 bits
```
themighty1, line 9: /// A block of 128 bits in big-endian byte order

mpz-core/src/cointoss/mod.rs
```diff
@@ -0,0 +1,43 @@
+//! A simple 2-party coin-toss protocol.
+//!
+//! # Example
```
themighty1, line 3: Does this example serve as a test? How can i run it? Does the CI need to be updated to also run examples?

themighty1 (approved):
gw, ack
i added tests which cover some corner cases

mpz-core/src/block.rs
sinui0: This comment would leak internal details of how the bits are stored. The user should not be privy to such things, ie assume it in their code.

mpz-core/src/cointoss/mod.rs
sinui0: it is a doctest, it runs when you run `cargo test`. The CI will run it as it is now

mpz-core/src/block.rs
themighty1: But a user who instantiates `let b = Block::new(bytes)` and later calls b.lsb() must know in which order to place their `bytes`, otherwise they will get  the wrong lsb, no?

themighty1: @sinui0 ---^

mpz-core/src/cointoss/mod.rs
themighty1: ok, I see. However, I noticed that vscode's code navigation doesnt work for doctests (lmk if it's just me). It creates unnecesary friction when examining the code. Can we duplicate the doctests also as normal tests? 
@sinui0

mpz-core/src/block.rs
sinui0: This is a good point. However, we should still avoid `endian` semantics, as that implies that the block is interpreted as an integer. I would opt to rename the `lsb` function to `color_bit` in this case. The user should not be privy to the location of the color bit in the underlying type.

mpz-core/src/cointoss/mod.rs
sinui0: I generally don't think code navigation should come into play for something like this. Ideally in the future we will have many many doctests, would we want to duplicate our entire test suite?

mpz-core/src/block.rs
themighty1: I agree with that direction. But using `color_bit` would leak the GC terminology. It can be called generically like e.g. `last_bit`or `rightmost_bit`

mpz-core/src/cointoss/mod.rs
themighty1: @sinui0 The first thing that I do when looking at how a protocol works is find an end-to-end test and from there to dive deep into it with code navigation. Not being able to do so would cause significant friction.

Until rust-analyzer supports navigation in doctests, we will have to duplicate those tests :(

PR #31, opened 2023-07-14
add warning to README

themighty1:
This PR adds a warning to prevent incorrect use of this lib.


th4s (approved):
ack

heeckhau (approved):
ack

sinui0:
Rather than imply that all protocols have this leakage/are meant to be revealed at the end, can we instead just give a more general warning about understanding the protocol before using it? This is a general purpose MPC lib, it just happens that we've implemented verifiable functionalities.

PR #36, opened 2023-07-20
Rewrite CO15

sinui0:
This PR rewrites CO15 and introduces changes to our OT traits. KOS has been removed and will be coming in a following PR.

# CO15 Changes
1. Added committed receiver functionality to CO15, which is required for committed sender in KOS.
2. Use simple incrementing nonce for the keys tweaks instead of merlin transcript. This is the same approach used by other implementations.
3. Typestates.
4. Introduced the async wrapper for CO15, previously it was coupled directly into KOS15 impl.
5. Toggle-able parallelism using rayon. This speeds up CO15 considerably, as the ECC ops can be fully parallelized.

# Interface changes
1. Renaming
2. We now use the `ProtocolMessage` trait to associate a message type with the implementing protocol.
3. Split the duplex channel interface to use both `Sink` and `Stream` instead. This provides a bit more flexibility when composing protocols together (this will be evident in the following KOS PR).
4. Introduced separate traits for different API requirements:
  - Trait for exclusive reference, where the implementing protocol does _not_ own a channel. This is what is implemented directly by the async wrappers and the lowest level API which provides the caller with control over IO.
  - Trait for shared reference, where the implementing protocol owns the IO channel. This is what our OT actor implementation will implement for the most abstracted and convenient way to use OT.
  - Trait for exclusive reference, where the implementing protocol owns the IO channel. This is mostly here for completeness, however I don't forsee an immediate requirement for it. We can likely provide an very thin wrapper which implements these traits for anything implementing the first set of traits.


ot/mpz-ot-core/Cargo.toml
```diff
@@ -19,12 +23,15 @@ cipher.workspace = true
...
 curve25519-dalek = { workspace = true, features = ["serde", "rand_core"] }
 serde = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 derive_builder.workspace = true
-merlin.workspace = true
-itybity.workspace = true
+itybity = { path = "../../../itybity" }
+opaque-debug.workspace = true
+cfg-if.workspace = true
+enum-try-as-inner = { tag = "0.1.0", git = "https://github.com/sinui0/enum-try-as-inner" }
```
th4s, line 34: Is this dependency on a private fork necessary? Any chance at doing a PR?

ot/mpz-ot-core/benches/ot.rs
```diff
@@ -1,71 +1,37 @@
...
-                let (mut receiver, receiver_setup) = receiver.extension_setup(&choice).unwrap();
-                let mut sender = sender
-                    .extension_setup(choice.len(), receiver_setup)
-                    .unwrap();
-
-                let send = sender.send(&msgs).unwrap();
-                let _received = receiver.receive(send).unwrap();
+                let receiver_payload = receiver.receive_random(choices.as_slice());
+                let sender_payload = sender.send(&msgs, receiver_payload).unwrap();
+                black_box(receiver.receive(sender_payload).unwrap())
```
th4s, line 24: Is there a reason why only this part is black-boxed?

Cargo.toml
```diff
@@ -23,7 +23,7 @@ clmul = { path = "clmul" }
 matrix-transpose = { path = "matrix-transpose" }
 
 tlsn-utils = { git = "https://github.com/tlsnotary/tlsn-utils" }
-tlsn-utils-aio = { git = "https://github.com/tlsnotary/tlsn-utils" }
+tlsn-utils-aio = { git = "https://github.com/tlsnotary/tlsn-utils", branch = "feature/duplex" }
```
th4s, line 26: Why does this depend on a different branch?

ot/mpz-ot-core/examples/ot.rs
```diff
@@ -1,46 +1,44 @@
...
-        [Block::new(14), Block::new(15)],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
+        [Block::ZERO, Block::ONES],
     ];
```
th4s: Let's replace it with:
```rust
    let inputs = [[Block::ZERO, Block::ONES]; 8];
```

ot/mpz-ot-core/src/chou_orlandi/error.rs
```diff
@@ -0,0 +1,43 @@
+/// Errors that can occur when using the CO15 sender.
+#[derive(Debug, thiserror::Error)]
+#[allow(missing_docs)]
+pub enum SenderError {
+    #[error("invalid state: expected {0}")]
+    InvalidState(String),
+    #[error("count mismatch: receiver expected {0} but sender sent {1}")]
```
th4s: I think it should be: `count mismatch: sender expected {0} but receiver sent {1}`

ot/mpz-ot-core/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,355 @@
...
+            ..
+        } = &mut self.state;
+
+        let ReceiverPayload { blinded_choices } = receiver_payload;
+
+        // Check that the number of inputs matches the number of choices
+        if inputs.len() != blinded_choices.len() {
+            return Err(SenderError::CountMismatch(
+                blinded_choices.len(),
+                inputs.len(),
```
th4s: Flip those two then.

ot/mpz-ot/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+};
+
+use crate::{OTError, OTReceiver, RevealChoices};
+
+use super::ReceiverError;
+
+#[derive(Debug, EnumTryAsInner)]
+enum State {
+    Initialized(Box<ReceiverCore<state::Initialized>>),
+    Setup(Box<ReceiverCore<state::Setup>>),
```
th4s: Why is this boxed here, but not for the Sender?

th4s (approved):
Nice work :rocket:!

ot/mpz-ot-core/Cargo.toml
sinui0: I'm going to ask the original author if he would accept a PR, but otherwise I will maintain the fork as I think there will be demand for it and it's quite lightweight.

ot/mpz-ot-core/benches/ot.rs
sinui0: its the end of the data dependency chain, all prior computation depends on it so the compiler shouldn't optimize it all away

Cargo.toml
sinui0: New IO traits, I need to open a PR in that repo

ot/mpz-ot-core/examples/ot.rs
sinui0: good call.

ot/mpz-ot/src/chou_orlandi/receiver.rs
sinui0: The Sender didn't have the allocation size disparity between the variants

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,282 @@
...
+        )
+    }
+}
+
+impl Receiver<state::Setup> {
+    /// Computes the decryption keys, returning the Receiver's payload to be sent to the Sender.
+    ///
+    /// # Arguments
+    ///
+    /// * `choices` - The receiver's choices
```
themighty1, line 105: Not all args have been described

ot/mpz-ot-core/src/chou_orlandi/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+pub use error::{ReceiverError, SenderError, SenderVerifyError};
+pub use receiver::{state as receiver_state, Receiver};
+pub use sender::{state as sender_state, Sender};
+
+use blake3::Hasher;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use mpz_core::Block;
+
+/// Hashes a ristretto point to a symmetric key
```
themighty1, line 21: ///
/// Prepending a tweak is suggested in [CO15] (section Non-Malleability in Practice).

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,282 @@
...
+            Self {
+                rng: ChaCha20Rng::from_entropy(),
+            }
+        }
+    }
+
+    /// The receiver's state after setup.
+    pub struct Setup {
+        pub(super) rng: ChaCha20Rng,
+        pub(super) sender_base_table: RistrettoBasepointTable,
```
themighty1, line 269: The "base table" terminology is Ristretto-specific, it shouldn't be leaked into our protocol. Can we have our own `Point` type to hide the "base table" complexity?

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn new_with_seed(config: ReceiverConfig, seed: [u8; 32]) -> Self {
+        Self {
+            config,
+            state: state::Initialized {
+                rng: ChaCha20Rng::from_seed(seed),
+            },
+        }
+    }
+
+    /// Sets up the receiver.
```
themighty1, line 79: /// Sets up the receiver. Returns the setup message to be sent to the sender.

themighty1 (changes requested):
gw, Im gonna quickly post an unfinished review, then when the coin-toss seed is implemented, will re-review again.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: Receiver setup message is now removed as the seed commitment occurs in the outer cointoss protocol.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: What argument is missing?

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: This is an additional abstraction that I think we should keep out of scope of this PR. The previous impl did not abstract this away

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
themighty1: sry, i was looking at the wrong place

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
themighty1: ok, i'll open a gh issue for that

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl Receiver {
+    /// Creates a new receiver.
+    ///
+    /// # Committed Receiver
+    ///
+    /// ## ⚠️ Warning ⚠️
+    ///
+    /// If the receiver is committed, the receiver's RNG seed must be unbiased such as generated by
+    /// a secure cointoss protocol with the sender.
```
themighty1: let's stick to the spelling of either "coin toss" (my preference) or "coin-toss" (already used by the cointoss mod)

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    iter.map(|(i, (b, c))| {
+        // blinded_choice is B in [ref1]
+        //
+        // if c = 0: B = g ^ b
+        // if c = 1: B = A * g ^ b
+        //
+        // when choice is 0, we add the zero element to keep constant time.
+        let blinded_choice = if c {
+            one + b * RISTRETTO_BASEPOINT_TABLE
```
themighty1: let's rename `one` to `A`, otherwise this line may cause cognitive friction

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+/// * `choices` - The choices of the OT receiver
+/// * `offset` - The number of OTs that have already been performed
+///              (used for the key derivation tweak)
+fn compute_decryption_keys<T: BitIterable + Sync>(
+    base_table: &RistrettoBasepointTable,
+    receiver_private_keys: &[Scalar],
+    choices: &[T],
+    offset: usize,
+) -> (Vec<RistrettoPoint>, Vec<(bool, Block)>) {
+    let zero = &Scalar::ZERO * base_table;
```
themighty1, line 197: I wasn't comfortable with these zero and one values, so I tested them. Can we put the tests below in this module for the time being. Later when we have a dedicated `Point` module we can move the tests there

```rust
#[cfg(test)]
mod tests {
    use rand_chacha::ChaCha12Rng;
    use rand_core::SeedableRng;

    use curve25519_dalek::{
        ristretto::{RistrettoBasepointTable, RistrettoPoint},
        scalar::Scalar,
    };

    #[test]
    fn test_multiply_base_table_by_0() {
        let mut rng = ChaCha12Rng::seed_from_u64(0);
        let point = RistrettoPoint::random(&mut rng);
        let table = RistrettoBasepointTable::create(&point);

        // an identity element, i.e. 0
        let identity = RistrettoPoint::default();
        assert_eq!(&Scalar::ZERO * &table, identity);
    }

    #[test]
    fn test_multiply_base_table_by_1() {
        let mut rng = ChaCha12Rng::seed_from_u64(1);
        let point = RistrettoPoint::random(&mut rng);
        let table = RistrettoBasepointTable::create(&point);

        assert_eq!(&Scalar::ONE * &table, point);
    }
}
```

ot/mpz-ot-core/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+}
+
+/// Computes the encryption keys for the sender.
+///
+/// # Arguments
+///
+/// * `private_key` - The sender's private key.
+/// * `public_key` - The sender's public key.
+/// * `receiver_payload` - The receiver's choice payload.
```
themighty1: should be `blinded_choices`

ot/mpz-ot-core/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,306 @@
...
+                blinded_choices.len(),
+            ));
+        }
+
+        if let Some(tape) = self.tape.as_mut() {
+            // Record the receiver's choices
+            tape.receiver_choices.extend_from_slice(&blinded_choices);
+        }
+
+        let mut payload =
```
themighty1, line 146: // Reusing `payload` to store the keys to avoid allocation.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        let (blinded_choices, decryption_keys) =
+            compute_decryption_keys(sender_base_table, &private_keys, choices, *counter);
+
+        *counter += blinded_choices.len();
+        cached_decryption_keys.extend(decryption_keys);
+
+        // If configured, log the choices
+        if self.config.receiver_commit() {
+            choice_log.extend(Vec::<u8>::from_lsb0_iter(choices.iter_lsb0()));
```
themighty1: if `choices.len()` is not a multiple of 8, then in case of multiple `receive_random` there will be padding zeroes in the log which will cause an error during Sender's verification.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+        impl Sealed for super::Initialized {}
+        impl Sealed for super::Setup {}
+    }
+
+    /// The receiver's state.
+    pub trait State: sealed::Sealed {}
+
+    /// The receiver's initial state.
+    pub struct Initialized {
+        pub(super) rng: ChaCha20Rng,
```
themighty1, line 249: /// RNG used to generate the receiver's keys

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+        fn default() -> Self {
+            Self {
+                rng: ChaCha20Rng::from_entropy(),
+            }
+        }
+    }
+
+    /// The receiver's state after setup.
+    pub struct Setup {
+        pub(super) rng: ChaCha20Rng,
```
themighty1, line 267: /// RNG used to generate the receiver's keys
        pub(super) rng: ChaCha20Rng,
        /// Sender's public key (precomputed table)
        pub(super) sender_base_table: RistrettoBasepointTable,
        /// Counts how many decryption keys we've computed so far
        pub(super) counter: usize,
        /// Log of the receiver's choice bits
        pub(super) choice_log: Vec<u8>,

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Computes the blinded choices `B` and the decryption keys for the OT receiver.
+///
+/// # Arguments
+///
+/// * `base_table` - A Ristretto basepoint table from the sender's public key
+/// * `receiver_private_keys` - The private keys of the OT receiver
+/// * `choices` - The choices of the OT receiver
+/// * `offset` - The number of OTs that have already been performed
```
themighty1: /// * `offset` - The number of decryption keys that have already been computed
///              (used for the key derivation tweak)

ot/mpz-ot/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// # Arguments
+    ///
+    /// * `sink` - The sink to send messages to the receiver
+    /// * `stream` - The stream to receive messages from the receiver
+    pub async fn setup<Si: IoSink<Message> + Send + Unpin, St: IoStream<Message> + Send + Unpin>(
+        &mut self,
+        sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<(), SenderError> {
+        let sender = self.state.replace(State::Error).into_initialized()?;
```
themighty1, line 78: the popup comment for replace is: "Replaces the current enum with the provided one, returning the old one"
this is confusing because we are replacing the enums' **variant**, not the enum itself.
Is this a comment from upstream or your fork? if the latter then could you pls update the comment.

ot/mpz-ot/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn from(value: enum_try_as_inner::Error<State>) -> Self {
+        SenderError::StateError(value.to_string())
+    }
+}
+
+/// Chou-Orlandi sender.
+#[derive(Debug)]
+pub struct Sender {
+    state: State,
+    cointoss_receiver: Option<cointoss::Receiver<cointoss::receiver_state::Received>>,
```
themighty1, line 38: /// The coin toss receiver after revealing one's own seed but before receiving a decommitment
    /// from the coin toss sender.

ot/mpz-ot/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        sink.send(Message::SenderSetup(msg)).await?;
+
+        self.state = State::Setup(sender);
+
+        Ok(())
+    }
+}
+
+async fn execute_cointoss<
```
themighty1, line 97: /// Executes the coin toss protocol as the receiver up until the point when the sender should send
/// a decommitment. The decommitment will be sent later during verification.

ot/mpz-ot/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        let receiver = Backend::spawn(move || receiver.setup(sender_setup)).await;
+
+        self.state = State::Setup(Box::new(receiver));
+
+        Ok(())
+    }
+}
+
+async fn execute_cointoss<
```
themighty1, line 117: /// Executes the coin toss protocol as the sender up until the point when we should send
/// a decommitment. The decommitment will be sent later during verification.

ot/mpz-ot/src/chou_orlandi/sender.rs
```diff
@@ -0,0 +1,201 @@
...
+        Si: IoSink<Message> + Send + Unpin,
+        St: IoStream<Message> + Send + Unpin,
+    >(
+        &mut self,
+        _sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<Vec<bool>, OTError> {
+        let sender = self
+            .state
+            .replace(State::Complete)
```
themighty1: this should be State::Error. State::Complete should be set at the end of this fn

themighty1 (approved):
gw!!!, the new EnumTryAsInner pattern feels very ergonomic.

ot/mpz-ot/src/chou_orlandi/receiver.rs
```diff
@@ -0,0 +1,226 @@
...
+        Si: IoSink<Self::Msg> + Send + Unpin,
+        St: IoStream<Self::Msg> + Send + Unpin,
+    >(
+        &mut self,
+        sink: &mut Si,
+        _stream: &mut St,
+    ) -> Result<(), OTError> {
+        let receiver = self
+            .state
+            .replace(State::Complete)
```
themighty1: should be State::Error here

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: uppercase variable names are reserved for constants, but I renamed to `a` and left a note.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: Testing dependencies like this is not a good practice. This is more of an API familiarity concern I think.

ot/mpz-ot-core/src/chou_orlandi/sender.rs
sinui0: Not sure I understand this, we're not reusing the receiver payload

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: Good catch. I'll change the storage to `Vec<bool>` for now, but at some point I may add `BitVec` to `itybity` or we can pull in `bitvec`

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
themighty1: Generally, yes, but when security is on the line, it doesn't hurt to be extra safe @sinui0

ot/mpz-ot-core/src/chou_orlandi/sender.rs
themighty1: More correctly this should be:
// Storing the keys in the `payload` variable to avoid allocation.
@sinui0

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: I'm not sure I understand, these tests are testing that the dependency is performing scalar multiplication correctly. That is an assumption we make when we add it as a dependency. We don't write tests for any of our other dependencies which could cause security issues, eg `std`

ot/mpz-ot-core/src/chou_orlandi/sender.rs
sinui0: I think this is more about naming, no? You're pointing out that we name the encryption keys `payload`, but its technically not the payload until we apply the keys to the messages. If you find this confusing maybe this would be better?

```rust
let keys = compute_encryption_keys(private_key, public_key, &blinded_choices, *counter);

// Encrypt the inputs
let mut payload = keys;
for (input, payload) in inputs.iter().zip(payload.iter_mut()) {
  payload[0] = input[0] ^ payload[0];
  payload[1] = input[1] ^ payload[1];
}
```

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
themighty1: @sinui0 , we also test that we correctly understand how multiplication by 0 and by 1 works for EC points.

This is not schoolbook math, so maybe * 0 does something different than what we think it does. Having those tests in place will also alleviate the concern of whoever is auditing the code.

**Generally** we trust our dependencies. But there is nothing wrong in being extra careful and testing some corner cases in case if our deps misbehave.

I actually ran into an issue with `rs-merkle` (https://github.com/antouhou/rs-merkle/pull/30) where the lib had a bug and did not sort the leaves. So, it is useful to test dependencies when in doubt, imo.

ot/mpz-ot-core/src/chou_orlandi/sender.rs
themighty1: yes, that's perfect, thanks.

ot/mpz-ot-core/src/chou_orlandi/receiver.rs
sinui0: > we also test that we correctly understand how multiplication by 0 and by 1 works for EC points

I do agree that is was good to check that, but now that we've verified it does what we expect I think we're good. That's what I meant about it being an API familiarity thing

> But there is nothing wrong in being extra careful and testing some corner cases in case if our deps misbehave.

I don't mean to come off like I'm arguing against security, but in this case I do think its not standard practice or good practice to include unit tests for a dependency like you're suggesting. Where do you draw the line of how far you test? There are many more test cases we would have to test in addition to the examples you've provided.

> I actually ran into an issue with rs-merkle

I think I would also categorize that as an API familiarity issue rather than a bug. You expected it to handle unsorted indices but it doesn't. Their docs should indicate that, or they should merge your changes. But that's a different class of issue compared to simply miscalculating the merkle root.

sinui0: This all being said, its not a hill i'll die on. If you feel strongly that we should include these tests I won't veto. Not sure the best place to put them, I feel like putting them in the CO15 module is odd

themighty1: @sinui0 , thanks, yes, I feel strongly about it :) Let's put them under receiver.rs and later we can move them into the `point` module when we have one.

PR #43, opened 2023-08-15
merge building blocks

xiangxiecrypto:
1. Add `reduce` function in `clmul`, it reduces the Galois Field polynomial. The current implementation supports x86/x86_64/aarch64 platforms, also a pure software implementation is provided.
2. Add a wrapper of AES encryption from RustCrypto, this forms the basis of other building blocks. The underlying AES implementation can be further improved, but I will leave it as a future work. For optimal performance, you should use the flags according to the instructions described in [RustCrypto AES](https://docs.rs/aes/latest/aes/).
3. Add `CR Hash` and  `CCR Hash`. 
4. Add `inner product` with or without `reduction` operations to `block.rs`.
5. Add `prg` (Pseudo Random Generator) based on AES.
6. Add implementation to generate `ggm_tree`. The implementation is based on `TwoKeyPrp`, which is based on AES.
7. Add benchmarks of `reduce`, `aes`, `prg` and `ggm`.


xiangxiecrypto:
Btw, what are the best flags should I use to get the optimal performance for `clmul`  on x86/x86_64. 
I used `RUSTFLAGS="-Ctarget-cpu=native" cargo bench --bench clmul`, but it is quite slow.

When I test on `ClmulX86`, the performance is much faster as expected.

sinui0:
> Btw, what are the best flags should I use to get the optimal performance for clmul

I'm not particularly familiar with the crate, but it shouldn't require any flags as it has runtime feature detection. From what I can see tho, `Clmul` is doing enum matching for every op which is likely slowing it down considerably. We can fix this in a separate PR.

mpz-core/src/aes.rs
```diff
@@ -42,4 +42,89 @@ impl FixedKeyAes {
...
+    }
+
+    /// Encrypt many blocks with many keys.
+    #[inline(always)]
+    pub fn para_encrypt<const NK: usize, const NM: usize>(keys: &[Self; NK], blks: &mut [Block]) {
+        let ptr = blks.as_mut_ptr() as *mut [Block; NM];
+        for i in 0..NK {
+            let ctxt = unsafe { &mut *ptr.add(i) };
+            *ctxt = keys[i].encrypt_many_blocks(*ctxt);
+        }
```
sinui0: This sort of pointer manipulation best be avoided.  I'm not sure I understand exactly what is going on here. What happens if `blks.len() < NK` or `blks.len() < NM`? This would lead to undefined behavior but is not warned in the docs nor is the function marked unsafe.

We should at least assert, but iterators can also be used as well:

```suggestion
        assert!(blks.len() >= NK + NM);
        let mut ctxt = [Block::default(); NM];
        keys.iter().enumerate().for_each(|(i, key)| {
            ctxt.copy_from_slice(&blks[i..NM + i]);
            blks[i..NM + i].copy_from_slice(&key.encrypt_many_blocks(ctxt))
        })
```

The above suggestion does not pass the unit test, so it may not be exactly what you had in mind but my point is to illustrate similar without using unsafe.

mpz-core/src/block.rs
```diff
@@ -58,6 +60,39 @@ impl Block {
...
+    /// The multiplication of two field elements.
+    #[inline]
+    pub fn gfmul(self, x: Self) -> Self {
+        let (a, b) = self.clmul(x);
+        Block::reduce(a, b)
+    }
+
+    /// Compute the inner product of two block vectors, without reducing the polynomial.
+    #[inline]
+    pub fn inn_prdt_no_red(a: &Vec<Block>, b: &Vec<Block>) -> (Block, Block) {
```
sinui0: Prefer `&[Block]` instead of `&Vec<Block>` as vectors implement `AsRef<[T]>` and the compiler automatically coerces this.

mpz-core/src/block.rs
```diff
@@ -69,6 +104,14 @@ impl Block {
     pub fn lsb(&self) -> usize {
         ((self.0[0] & 1) == 1) as usize
     }
+
+    ///Function ``sigma( x0 || x1 ) = (x0 xor x1) || x1``.
```
sinui0: Missing space `/// Function...`

mpz-core/src/block.rs
```diff
@@ -155,12 +205,25 @@ impl BitAnd for Block {
...
         Block::new(rng.gen())
     }
 }
 
+impl AsMut<[u8]> for Block {
+    #[inline(always)]
+    fn as_mut(&mut self) -> &mut [u8] {
+        unsafe { &mut *(self as *mut Block as *mut [u8; 16]) }
+    }
+}
```
sinui0, line 235: Add newline below

mpz-core/src/block.rs
```diff
@@ -155,12 +205,25 @@ impl BitAnd for Block {
...
 impl Distribution<Block> for Standard {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Block {
         Block::new(rng.gen())
     }
 }
 
+impl AsMut<[u8]> for Block {
+    #[inline(always)]
+    fn as_mut(&mut self) -> &mut [u8] {
+        unsafe { &mut *(self as *mut Block as *mut [u8; 16]) }
```
sinui0:
```suggestion
        self.0.as_mut()
```

mpz-core/src/ggm_tree.rs
```diff
@@ -0,0 +1,54 @@
+//! Implement GGM tree for OT.
```
sinui0, line 1: Can we add a reference for this construction for those unfamiliar (me :sweat_smile: )

mpz-core/src/ggm_tree.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub fn new(depth: usize) -> Self {
+        let tkprp = TwoKeyPrp::new([Block::ZERO, Block::from(1u128.to_le_bytes())]);
+        Self { tkprp, depth }
+    }
+
+    /// Generate tree.\
+    /// Take as input a `seed`. \
+    /// The generated tree is stored in `tree`. \
+    /// Output two block slices used for OT.
+    /// This implementation is adopted from EMP Toolkit.
```
sinui0: Please add newline instead of `\` and document the function arguments

mpz-core/src/ggm_tree.rs
sinui0, line 1: Can we please get some unit tests for this, even if just against a couple test vectors :pray:

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[inline(always)]
+    fn generate(&mut self, results: &mut Self::Results) {
+        let states = [0; AesEncryptor::AES_BLOCK_SIZE].map(
+            #[inline(always)]
+            |_| {
+                let x = self.state;
+                self.state += 1;
+                Block::from(unsafe { mem::transmute::<_, [u8; 16]>([x, 0u64]) })
+            },
+        );
```
sinui0: We can avoid unsafe:

```suggestion
        let states: [Block; AesEncryptor::AES_BLOCK_SIZE] = 
            std::array::from_fn(|i| (i as u128).to_le_bytes().into()));
        self.state += AesEncryptor::AES_BLOCK_SIZE as u64;
```

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,154 @@
...
+        *results = unsafe { mem::transmute(self.aes.encrypt_many_blocks(states)) }
```
sinui0: This sort of memory operation makes an assumption about the underlying storage of a `Block`, coupling the `prg` module to such details. Can we instead encapsulate this into a helper function within the `block` module?

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[inline(always)]
+    pub fn random_block(&mut self) -> Block {
+        self.gen()
+    }
+
+    /// Fill a block slice with random block values.
+    #[inline(always)]
+    pub fn random_blocks(&mut self, buf: &mut [Block]) {
+        let bytes =
+            unsafe { core::slice::from_raw_parts_mut(buf.as_ptr() as *mut u8, buf.len() * 16) };
```
sinui0: This is unsound as it assumes that `buf` is a [single allocated object](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#incorrect-usage) and it makes an assumption about the underlying memory layout of `Block`

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[inline(always)]
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[test]
+fn prg_test() {
+    let mut prg = Prg::new();
+    // let x: bool = prg.gen();
```
sinui0: dead code

mpz-core/src/tkprp.rs
```diff
@@ -0,0 +1,87 @@
+//! Implement the two-key PRG as G(k) = PRF_seed0(k)\xor k || PRF_seed1(k)\xor k
```
sinui0, line 1: Please provide a reference for this construction if it exists

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,154 @@
...
+use rand::Rng;
+use rand_core::{
+    block::{BlockRng, BlockRngCore},
+    CryptoRng, RngCore, SeedableRng,
+};
+use std::mem;
+
+///Struct of PRG Core
+#[derive(Clone)]
+pub struct PrgCore {
```
sinui0: Does this need to be `pub`?

sinui0 (changes requested):
Good work, thank you for the PR :pray:

I mostly have some requests around the usage of `unsafe`. Besides that, please be sure to run `rustfmt` and `clippy`. Additionally, we will add [`miri`](https://github.com/rust-lang/miri) testing to our CI to accompany any new unsafe code, which can be run locally as well. Our CI didn't run to catch this because this PR isn't against our main branch, I will try to enable it sometime tomorrow.

For any additions of unsafe code, please make sure to document any invariants required by each `unsafe` block. If the function is marked safe but contains unsafe code, please provide rationale for why such invariants will be upheld in all circumstances, eg prior assertions.

Our bias in this codebase is safety > performance. That isn't to say _never_ use unsafe, but there is an additional burden on maintainability, review and safety. We typically follow "make it work, then make it fast", and most things can be implemented using safe Rust. Later we can use profiling to identify pieces which are worth optimizing

xiangxiecrypto:
Good point in handling `unsafe`, the codebase is still in its early stage, a good chance to make it robust.

xiangxiecrypto:
> > Btw, what are the best flags should I use to get the optimal performance for clmul
> 
> I'm not particularly familiar with the crate, but it shouldn't require any flags as it has runtime feature detection. From what I can see tho, `Clmul` is doing enum matching for every op which is likely slowing it down considerably. We can fix this in a separate PR.

I agree, the `match` branch is dominating the cost of `Clmul`. In an AWS `m5.xlarge` machine, the performance of  clmul is about 47 ns, while in `ClmulX86`, it only needs 0.8 ns.

mpz-core/src/prg.rs
sinui0: Just noticed this is wrong as we're not adding `self.state` to `i`, but the same jist

sinui0:
Opened #44 for improving `clmul`

mpz-core/src/aes.rs
xiangxiecrypto: updated without using `unsafe`

mpz-core/src/block.rs
xiangxiecrypto: Done

mpz-core/src/block.rs
xiangxiecrypto: Done

mpz-core/src/block.rs
xiangxiecrypto: Done

mpz-core/src/block.rs
xiangxiecrypto: Done

mpz-core/src/ggm_tree.rs
xiangxiecrypto: Done

mpz-core/src/ggm_tree.rs
xiangxiecrypto: Done

mpz-core/src/ggm_tree.rs
xiangxiecrypto: Done

mpz-core/src/prg.rs
xiangxiecrypto: Remove all the above unsafe part by using `bytemuck`

mpz-core/src/prg.rs
xiangxiecrypto: replace the unsafe part with `bytemuck`

mpz-core/src/prg.rs
xiangxiecrypto: removed

mpz-core/src/tkprp.rs
xiangxiecrypto: Done

mpz-core/src/prg.rs
xiangxiecrypto: yes, the standard implementation needs it be `pub`

xiangxiecrypto:
Basically, removing all the `unsafe` by using `bytemuck`.

clmul/src/backend.rs
```diff
@@ -9,6 +9,37 @@ use core::ops::{BitXor, BitXorAssign};
...
+        let a = x3 >> 63;
+        let b = x3 >> 62;
+        let c = x3 >> 57;
+        let d = x2 ^ a ^ b ^ c;
+        let (e1, e0) = sep(join(x3, d) << 1);
+        let (f1, f0) = sep(join(x3, d) << 2);
+        let (g1, g0) = sep(join(x3, d) << 7);
+        let h1 = x3 ^ e1 ^ f1 ^ g1;
+        let h0 = d ^ e0 ^ f0 ^ g0;
+        unsafe { core::mem::transmute(join(x1 ^ h1, x0 ^ h0)) }
```
sinui0: can use bytemuck here

mpz-core/src/block.rs
sinui0: Maybe it didnt get included in commits?

mpz-core/src/prg.rs
sinui0: Can you elaborate on this? What outside of this module will be using `PrgCore`?

mpz-core/src/block.rs
```diff
@@ -69,6 +105,14 @@ impl Block {
     pub fn lsb(&self) -> usize {
         ((self.0[0] & 1) == 1) as usize
     }
+
+    ///Function ``sigma( x0 || x1 ) = (x0 xor x1) || x1``.
+    #[inline(always)]
+    pub fn sigma(a: Self) -> Self {
+        let mut x: [u64; 2] = unsafe { std::mem::transmute(a) };
+        x[0] ^= x[1];
+        Block(unsafe { std::mem::transmute(x) })
```
sinui0: bytemuck

sinui0 (approved):
Goodwork, thanks for the changes.

I've requested a couple more small changes, but otherwise, approved!

clmul/Cargo.toml
```diff
@@ -10,6 +10,8 @@ cfg-if.workspace = true
 [target.'cfg(any(target_arch = "aarch64", target_arch = "x86_64", target_arch = "x86"))'.dependencies]
 cpufeatures.workspace = true
 
+bytemuck = {workspace = true, features = ["derive"]}
```
sinui0: this only adds `bytemuck` for certain targets, move it into the `dependencies` table

xiangxiecrypto:
updated accordingly.

clmul/src/backend.rs
```diff
@@ -9,6 +9,37 @@ use core::ops::{BitXor, BitXorAssign};
 #[cfg_attr(target_pointer_width = "64", path = "backend/soft64.rs")]
 mod soft;
 
+impl soft::Clmul {
+    pub fn reduce(x: Self, y: Self) -> Self {
```
themighty1: Since the `clmul` crate is generic over the polynomial, let's call this `reduce_gcm`. 
Can we put a comment above the `fn`
   /// Reduces the polynomial represented in bits modulo the GCM polynomial x^128 + x^7 + x^2 + x + 1.
    /// `x` and `y` are resp. upper and lower bits of the polynomial.
    ///
    /// Page 16 of [Intel® Carry-Less Multiplication Instruction and its Usage for Computing the GCM Mode rev 2.02]
    /// (https://www.intel.com/content/dam/develop/external/us/en/documents/clmul-wp-rev-2-02-2014-04-20.pdf)

clmul/src/backend.rs
```diff
@@ -150,6 +181,41 @@ impl Clmul {
             },
         }
     }
+
+    /// Performs reduction
+    pub fn reduce(x: Self, y: Self) -> Self {
```
themighty1: This can also be called `reduce_gcm` with a comment:
/// Reduces the polynomial represented in bits modulo the GCM polynomial x^128 + x^7 + x^2 + x + 1.
/// `x` and `y` are resp. upper and lower bits of the polynomial.

clmul/src/backend.rs
```diff
@@ -232,3 +298,27 @@ impl PartialEq for Clmul {
         }
     }
 }
+
+#[test]
+fn reduce_test() {
+    use rand::Rng;
+    use rand_chacha::{rand_core::SeedableRng, ChaCha12Rng};
+
+    let mut rng = ChaCha12Rng::from_entropy();
```
themighty1: we have a convention of using reproducible randomness in tests: `ChaCha12Rng::from_seed([0; 32]);`
pls lmk if you think `from_entropy()` is preferable

clmul/src/backend.rs
```diff
@@ -150,6 +181,41 @@ impl Clmul {
...
     }
+
+    /// Performs reduction
+    pub fn reduce(x: Self, y: Self) -> Self {
+        match x.intrinsics {
+            Some(x_intr) => match y.intrinsics {
+                Some(y_intr) => {
+                    cfg_if! {
+                        if #[cfg(any(all(target_arch = "aarch64", feature = "armv8"), any(target_arch = "x86_64", target_arch = "x86")))]{
+                            let r = intrinsics::Clmul::reduce(x_intr, y_intr);
```
themighty1: if Im not mistaken, this will call `reduce_unsafe` which has
`#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]`
 which means that there will be a compilation error when compiling on `all(target_arch = "aarch64", feature = "armv8")`.
if so, let's just remove `all(target_arch = "aarch64", feature = "armv8")`

clmul/src/backend/clmul.rs
```diff
@@ -84,6 +84,42 @@ impl ClmulX86 {
             ClmulX86(_mm_unpacklo_epi64(v2, v3)),
         )
     }
+
+    #[inline(always)]
+    pub fn reduce(x: Self, y: Self) -> ClmulX86 {
```
themighty1: `reduce_gcm`

clmul/src/backend/clmul.rs
```diff
@@ -84,6 +84,42 @@ impl ClmulX86 {
...
+        unsafe { Self::reduce_unsafe(&x, &y) }
+    }
+
+    // This implementation is adapted from EMP Toolkit.
+    #[inline]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    unsafe fn reduce_unsafe(x: &Self, y: &Self) -> ClmulX86 {
```
themighty1: `reduce_gcm_unsafe`

clmul/src/backend/pmull.rs
```diff
@@ -83,6 +83,63 @@ impl ClmulArm {
 
         (ClmulArm(r0), ClmulArm(r1))
     }
+
+    #[inline(always)]
+    pub fn reduce(x: Self, y: Self) -> ClmulArm {
```
themighty1: `reduce_gcm`

clmul/src/backend/pmull.rs
```diff
@@ -83,6 +83,63 @@ impl ClmulArm {
...
+        unsafe { Self::reduce_unsafe(&x, &y) }
+    }
+
+    // This implementation is adapted from EMP Toolkit.
+    #[inline]
+    #[target_feature(enable = "neon")]
+    unsafe fn reduce_unsafe(x: &Self, y: &Self) -> ClmulArm {
```
themighty1: `reduce_gcm_unsafe`

clmul/src/backend/pmull.rs
```diff
@@ -83,6 +83,63 @@ impl ClmulArm {
...
     }
+
+    #[inline(always)]
+    pub fn reduce(x: Self, y: Self) -> ClmulArm {
+        unsafe { Self::reduce_unsafe(&x, &y) }
+    }
+
+    // This implementation is adapted from EMP Toolkit.
+    #[inline]
+    #[target_feature(enable = "neon")]
```
themighty1, line 94: @xiangxiecrypto , were you able to test this code on a CPU with NEON?

mpz-core/src/block.rs
```diff
@@ -69,6 +105,14 @@ impl Block {
     pub fn lsb(&self) -> usize {
         ((self.0[0] & 1) == 1) as usize
     }
+
+    /// Function ``sigma( x0 || x1 ) = (x0 xor x1) || x1``.
```
themighty1: If Im not mistaken, the paper says ` σ(xL || xR ) = xR ⊕ xL || xL `
there's `xL` at the end, so it should be x0 in our case

mpz-core/src/aes.rs
```diff
@@ -42,4 +42,93 @@ impl FixedKeyAes {
...
+        let mut x = GenericArray::from(block);
+        self.aes.encrypt_block(&mut x);
+        Block::from(x) ^ block
+    }
+
+    /// Circular correlation-robust hash function
+    /// (cf.<https://eprint.iacr.org/2019/074>, §7.3).
+    ///
+    /// The function computes `H(sigma(x))`, where `H` is a correlation-robust hash
+    /// function and `sigma( x0 || x1 ) = (x0 xor x1) || x1`.
```
themighty1: `x0` at the end (see my comment to `sigma`)

mpz-core/src/aes.rs
```diff
@@ -42,4 +42,93 @@ impl FixedKeyAes {
...
+        let mut ctxt = [Block::default(); N];
+        for i in 0..N {
+            ctxt[i] = self.encrypt_block(blks[i]);
+        }
+        ctxt
+    }
+
+    /// Encrypt many blocks with many keys.
+    /// Input: `NK` AES keys `keys`, and `NK * NM` blocks `blks`
+    /// Output: use each AES key encrypts each bunch of `NM` blocks.
```
themighty1: fixing some grammar:
/// Output: each batch of `NM` blocks encrypted by a corresponding AES key.
/// Only handle the first `NK * NM` blocks of `blks`, do not handle the rest.

mpz-core/src/aes.rs
```diff
@@ -42,4 +42,93 @@ impl FixedKeyAes {
...
+    let aes1 = AesEncryptor::new(Block::ONES);
+
+    let mut blks = [Block::default(); 4];
+    blks[1] = Block::ONES;
+    blks[3] = Block::ONES;
+    AesEncryptor::para_encrypt::<2, 2>(&[aes, aes1], &mut blks);
+    assert_eq!(
+        blks,
+        [
+            Block::from((0x2E2B34CA59FA4C883B2C8AEFD44BE966 as u128).to_le_bytes()),
```
themighty1: clippy expects `0x4E668D3ED24773FA0A5A85EAC98C5B3F_u128`

mpz-core/src/block.rs
```diff
@@ -58,6 +61,39 @@ impl Block {
         (Self::new(a.into()), Self::new(b.into()))
     }
 
+    #[inline]
+    /// Performs reduction
```
themighty1: /// Reduces the polynomial represented in bits modulo the GCM polynomial x^128 + x^7 + x^2 + x + 1.
    /// `x` and `y` are resp. upper and lower bits of the polynomial.

mpz-core/src/block.rs
```diff
@@ -58,6 +61,39 @@ impl Block {
...
+    pub fn reduce(x: Self, y: Self) -> Self {
+        let r = Clmul::reduce(Clmul::new(&x.0), Clmul::new(&y.0));
+        Self::new(r.into())
+    }
+
+    /// The multiplication of two field elements.
```
themighty1: /// The multiplication of two Galois field elements.

mpz-core/src/block.rs
```diff
@@ -58,6 +61,39 @@ impl Block {
...
+        Block::reduce(a, b)
+    }
+
+    /// Compute the inner product of two block vectors, without reducing the polynomial.
+    #[inline]
+    pub fn inn_prdt_no_red(a: &[Block], b: &[Block]) -> (Block, Block) {
+        assert_eq!(a.len(), b.len());
+        a.iter()
+            .zip(b.iter())
+            .fold((Block::default(), Block::default()), |acc, (x, y)| {
```
themighty1: let's use `Block::ZERO` here for expliciteness

mpz-core/src/block.rs
```diff
@@ -206,4 +271,33 @@ mod tests {
...
+        let mut c = (Block::default(), Block::default());
+        let mut d = Block::default();
+        for i in 0..SIZE {
+            let r: [u8; 16] = rng.gen();
+            a.push(Block::from(r));
+            let r: [u8; 16] = rng.gen();
+            b.push(Block::from(r));
+
+            let z = a[i].clmul(b[i]);
+            c.0 = c.0 ^ z.0;
```
themighty1, line 313: clippy warnings here

mpz-core/src/block.rs
```diff
@@ -206,4 +271,33 @@ mod tests {
         let a = Block::new(three);
         assert_eq!(a.lsb(), 1);
     }
+
+    #[test]
+    fn inn_prdt_test() {
+        use rand::{Rng, SeedableRng};
+        use rand_chacha::ChaCha12Rng;
+        let mut rng = ChaCha12Rng::from_entropy();
```
themighty1: ChaCha12Rng::from_seed([0; 32]) is preferable

mpz-core/src/aes.rs
```diff
@@ -42,4 +42,93 @@ impl FixedKeyAes {
...
+    }
+}
+
+/// A wrapper of aes, only for encryption.
+#[derive(Clone)]
+pub struct AesEncryptor(Aes128Enc);
+
+impl AesEncryptor {
+    /// AES_BLOCK_SIZE
+    pub const AES_BLOCK_SIZE: usize = 8;
```
themighty1: this var seems `prg.rs` specific. maybe it should live there?

mpz-core/src/prg.rs
```diff
@@ -0,0 +1,149 @@
...
+    #[inline(always)]
+    fn from_rng<R: RngCore>(rng: R) -> Result<Self, rand_core::Error> {
+        BlockRng::<PrgCore>::from_rng(rng).map(Prg)
+    }
+}
+
+impl CryptoRng for Prg {}
+
+impl Prg {
+    /// New a Prg with random seed.
```
themighty1: /// New Prg with random seed.

themighty1 (changes requested):
Thank you, looks good. Will need to address some comments.

clmul/src/backend.rs
xiangxiecrypto: This follows the structure of the `new` function, and I did not get a compilation error on an `aarch64` platform with `armv8` enabled.

clmul/src/backend/pmull.rs
xiangxiecrypto: yes, on my M1 MAC (without Rosetta installed).

mpz-core/src/block.rs
xiangxiecrypto: I should clarify in the comment. `x0` (`xR`) is the lower half of `x`, `x1` (`xL`) is the higher half of `x`

mpz-core/src/block.rs
xiangxiecrypto: a little bit weird that I did not get the warning.

mpz-core/src/aes.rs
xiangxiecrypto: It will be used in other places, not just `Prg`, and it is related to `AES`.

xiangxiecrypto:
Thank you @themighty1 for the careful review, and changed it accordingly.

clmul/src/backend.rs
themighty1: yes, sry, I forgot that on arm it will call `reduce_unsafe` from `pmull.rs`

mpz-core/src/aes.rs
themighty1: im ok to keep it here, can we maybe call it `AES_BLOCK_COUNT` and give it a more descriptive comment?

themighty1: @xiangxiecrypto ---^

mpz-core/src/block.rs
themighty1: @xiangxiecrypto, I think the code for sigma is incorrect. The paper says that 
1) the first half must xored to the second half
2) the second half must be replaced with the first half
The current code only performs 1) and does not perform 2)

xiangxiecrypto: Oops. Yes, I think you are right. I should add a test for `sigma`.

themighty1 (approved):
Thank you, lgtm

sinui0:
Thanks @themighty1 and @xiangxiecrypto 

I will merge this PR soon, it might have conflicts with a couple other branches in queue. I'll fix the conflicts afterwards.

xiangxiecrypto:
> Thanks @themighty1 and @xiangxiecrypto
> 
> I will merge this PR soon, it might have conflicts with a couple other branches in queue. I'll fix the conflicts afterwards.

Cheers!

sinui0:
@xiangxiecrypto we're ready to merge this. I don't think I have permissions for your branch, if you don't mind fixing the conflict here I will merge it afterwards

xiangxiecrypto:
> 

conflicts resolved.

PR #48, opened 2023-08-20
KOS arbitrary message length

sinui0:
This feature was forgotten in the rewrite but is required for our share conversion protocols (P256 element is 32 bytes).

I implemented this a bit differently than it was previously, this time supporting arbitrary _byte_ length instead of _block_ length using AES-CTR.


ot/mpz-ot-core/src/kos/receiver.rs
```diff
@@ -469,6 +474,59 @@ impl ReceiverKeys {
...
+            return Err(ReceiverError::IdMismatch(self.id, id));
+        }
+
+        let length = length as usize;
+        if ciphertexts.len() / (2 * length) != self.keys.len() {
+            return Err(ReceiverError::CountMismatch(
+                self.keys.len(),
+                ciphertexts.len() / (2 * length),
+            ));
+        }
```
th4s, line 515: I do not get why you have to divide by `2 * length`?

ot/mpz-ot-core/src/kos/sender.rs
```diff
@@ -381,7 +382,83 @@ impl SenderKeys {
...
+                    let mut m1 = *m1;
+
+                    e0.apply_keystream(&mut m0);
+                    e1.apply_keystream(&mut m1);
+
+                    [m0, m1]
+                })
+                .flatten()
+                .collect()
+        };
```
th4s: I think we could simplify to
```rust
        // If we have derandomization, use it to correct the receiver's choices, else we use
        // default
        let flip = self
            .derandomize
            .map(|x| x.flip)
            .unwrap_or_else(|| vec![0_u8; self.keys.len() / 8 + 1]);

        // Generate a random IV which is used for all messages.
        // This is safe because every message is encrypted with a different key.
        let iv: [u8; 16] = rand::thread_rng().gen();

        // Encrypt the chosen messages using the generated keys from ROT.
        let ciphertexts = self.keys
                .into_iter()
                .zip(msgs)
                .zip(flip.iter_lsb0())
                .flat_map(|(([k0, k1], [m0, m1]), flip)| {
                    // Initialize AES-CTR with the keys from ROT.
                    let mut e0 = Aes128Ctr::new(&k0.into(), &iv.into());
                    let mut e1 = Aes128Ctr::new(&k1.into(), &iv.into());

                    let mut m0 = *m0;
                    let mut m1 = *m1;

                    // Use Beaver derandomization to correct the receiver's choices
                    // from the extension phase.
                    if flip {
                        e1.apply_keystream(&mut m0);
                        e0.apply_keystream(&mut m1);
                    } else {
                        e0.apply_keystream(&mut m0);
                        e1.apply_keystream(&mut m1);
                    }

                    [m0, m1]
                })
                .flatten()
                .collect();
```

ot/mpz-ot-core/src/kos/sender.rs
```diff
@@ -381,7 +382,83 @@ impl SenderKeys {
...
+        // Generate a random IV which is used for all messages.
+        // This is safe because every message is encrypted with a different key.
+        let iv: [u8; 16] = rand::thread_rng().gen();
+
+        // Encrypt the chosen messages using the generated keys from ROT.
+        let ciphertexts = if let Some(Derandomize { flip, .. }) = self.derandomize {
+            self.keys
+                .into_iter()
+                .zip(msgs)
+                .zip(flip.iter_lsb0())
```
th4s: Shouldn't we also check that the length of `flip` is what we expect?

th4s (approved):
Nice :rocket:, some questions/comments

ot/mpz-ot-core/src/kos/receiver.rs
sinui0: There are 2 ciphertexts per OT, and the ciphertexts are `length` long

ot/mpz-ot-core/src/kos/sender.rs
sinui0: This happens here:

https://github.com/privacy-scaling-explorations/mpz/blob/b41b2bd0400c2ae22c586c670eebf148c9ef5bfe/ot/mpz-ot-core/src/kos/sender.rs#L332-L347

sinui0: Oh I see why it's not totally clear. `flip` length is checked to be consistent during deserialization:

https://github.com/privacy-scaling-explorations/mpz/blob/b41b2bd0400c2ae22c586c670eebf148c9ef5bfe/ot/mpz-ot-core/src/msgs.rs#L25-L45

ot/mpz-ot-core/src/kos/receiver.rs
th4s: But don''t we use a single key per ciphertext? I thought that was the idea of having OT with arbitrary byte length.

sinui0: `self.keys` is `Vec<[Block; 2]>`

th4s: I see it now :+1: Makes sense

ot/mpz-ot-core/src/kos/receiver.rs
```diff
@@ -469,6 +474,59 @@ impl ReceiverKeys {
...
+            .into_iter()
+            .zip(self.choices)
+            .zip(ciphertexts.chunks(2 * length))
+            .map(|((key, c), ct)| {
+                // Initialize AES-CTR with the key from ROT.
+                let mut e = Aes128Ctr::new(&key.into(), &iv.into());
+
+                let mut msg = [0u8; N];
+                if c {
+                    msg.copy_from_slice(&ct[length..])
```
themighty1: This assumes that `N == length` but I don't see where this equality was checked before.
Also, we don't seems to need to get `length` from the other party if we pass `N` in ourselves.

ot/mpz-ot-core/src/kos/receiver.rs
```diff
@@ -469,6 +474,59 @@ impl ReceiverKeys {
...
+
+        let length = length as usize;
+        if ciphertexts.len() / (2 * length) != self.keys.len() {
+            return Err(ReceiverError::CountMismatch(
+                self.keys.len(),
+                ciphertexts.len() / (2 * length),
+            ));
+        }
+
+        let iv: [u8; 16] = iv.try_into().map_err(|_| ReceiverError::InvalidPayload)?;
```
themighty1: This is a CTR nonce which must be only 8 bytes. I don't understand why `Aes128Ctr::new` below silently accepts it without giving an error.

themighty1 (approved):
gw! lgtm modulo the comments

ot/mpz-ot-core/src/kos/receiver.rs
sinui0: It is an IV, which gets combined with the CTR nonce. The CTR nonce is per block, the IV is per message

ot/mpz-ot-core/src/kos/receiver.rs
sinui0: Good catch.

I've added an equality check. I kept the `length` field in `SenderPayload` for better error reporting

PR #52, opened 2023-08-22
OTSetup trait

sinui0:
This PR introduces the `OTSetup` trait for performing any one-time setup that needs to be done before using other functionalities. One day I would like this to be a more general trait for any protocol (not just OT), but this will work for now.

This saves us from having to set up base OT before injecting it into OT extension.

### DIFF
--- a/ot/mpz-ot/src/kos/receiver.rs
+++ b/ot/mpz-ot/src/kos/receiver.rs
@@ -164,11 +164,6 @@ where
             .into_initialized()
             .map_err(ReceiverError::from)?;
 
-        // Set up base OT if not already done
-        self.base
-            .setup(&mut into_base_sink(sink), &mut into_base_stream(stream))
-            .await?;
-
         // If the sender is committed, we run a cointoss
         if ext_receiver.config().sender_commit() {
             let commitment = stream
@@ -185,6 +180,11 @@ where
 
             self.cointoss_receiver = Some(cointoss_receiver);
         }
+
+        // Set up base OT if not already done
+        self.base
+            .setup(&mut into_base_sink(sink), &mut into_base_stream(stream))
+            .await?;
 
         let seeds: [[Block; 2]; CSP] = std::array::from_fn(|_| thread_rng().gen());
 
--- a/ot/mpz-ot/src/kos/receiver.rs
+++ b/ot/mpz-ot/src/kos/receiver.rs
@@ -14,7 +14,7 @@ use utils_aio::{
     stream::{ExpectStreamExt, IoStream},
 };
 
-use crate::{OTError, OTReceiver, OTSender, VerifiableOTReceiver, VerifiableOTSender};
+use crate::{OTError, OTReceiver, OTSender, OTSetup, VerifiableOTReceiver, VerifiableOTSender};
 
 use super::{into_base_sink, into_base_stream, ReceiverError, ReceiverVerifyError};
 
@@ -63,53 +63,6 @@ where
     /// The number of remaining OTs which can be consumed.
     pub fn remaining(&self) -> Result<usize, ReceiverError> {
         Ok(self.state.as_extension()?.remaining())
-    }
-
-    /// Performs the base OT setup.
-    ///
-    /// # Arguments
-    ///
-    /// * `sink` - The sink to send messages to the base OT receiver
-    /// * `stream` - The stream to receive messages from the base OT receiver
-    pub async fn setup<
-        Si: IoSink<Message<BaseOT::Msg>> + Send + Unpin,
-        St: IoStream<Message<BaseOT::Msg>> + Send + Unpin,
-    >(
-        &mut self,
-        sink: &mut Si,
-        stream: &mut St,
-    ) -> Result<(), ReceiverError> {
-        let ext_receiver = self.state.replace(State::Error).into_initialized()?;
-
-        // If the sender is committed, we run a cointoss
-        if ext_receiver.config().sender_commit() {
-            let commitment = stream.expect_next().await?.into_cointoss_commit()?;
-
-            let (cointoss_receiver, payload) =
-                cointoss::Receiver::new(vec![thread_rng().gen()]).reveal(commitment)?;
-
-            sink.send(Message::CointossReceiverPayload(payload)).await?;
-
-            self.cointoss_receiver = Some(cointoss_receiver);
-        }
-
-        let mut rng = thread_rng();
-        let seeds: [[Block; 2]; CSP] = std::array::from_fn(|_| [rng.gen(), rng.gen()]);
-
-        // Send seeds to sender
-        self.base
-            .send(
-                &mut into_base_sink(sink),
-                &mut into_base_stream(stream),
-                &seeds,
-            )
-            .await?;
-
-        let ext_receiver = ext_receiver.setup(seeds);
-
-        self.state = State::Extension(Box::new(ext_receiver));
-
-        Ok(())
     }
 
     /// Performs OT extension.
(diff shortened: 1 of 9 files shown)


ot/mpz-ot/src/kos/receiver.rs
```diff
@@ -188,6 +141,70 @@ where
...
+    async fn setup<
+        Si: IoSink<Message<BaseOT::Msg>> + Send + Unpin,
+        St: IoStream<Message<BaseOT::Msg>> + Send + Unpin,
+    >(
+        &mut self,
+        sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<(), OTError> {
+        if self.state.is_extension() {
+            return Ok(());
```
themighty1, line 158: This will allow to re-use the same delta which was already used in the previous KOS extension?
I think it is not safe to do so, because after the first KOS extension, some bits of delta may have been leaked.
If we re-use a leaky delta in a second KOS extension, then the security analysis of the KOS paper doesn't apply anymore, because the paper assumes that the KOS extension starts with a delta which hasn't yet been leaked.

The safe option of doing this would be as per https://crypto.stackexchange.com/questions/101310/reusing-base-oblivious-transfers 
to allocate 128 extra OTs from the first KOS extension and use them for the secons KOS extension.

ot/mpz-ot/src/kos/receiver.rs
```diff
@@ -188,6 +141,70 @@ where
...
+        }
+
+        let ext_receiver = self
+            .state
+            .replace(State::Error)
+            .into_initialized()
+            .map_err(ReceiverError::from)?;
+
+        // Set up base OT if not already done
```
themighty1: since we cannot be here if there already was a base OT setup, we can remove `if not already done`

ot/mpz-ot/src/kos/sender.rs
```diff
@@ -233,6 +196,64 @@ where
...
+    >(
+        &mut self,
+        sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<(), OTError> {
+        if self.state.is_extension() {
+            return Ok(());
+        }
+
+        // Set up base OT if not already done
```
themighty1: the base ot setup will always be "not done" when we get here.

ot/mpz-ot/src/kos/sender.rs
```diff
@@ -233,6 +196,64 @@ where
...
+        St: IoStream<Message<BaseOT::Msg>> + Send + Unpin,
+    >(
+        &mut self,
+        sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<(), OTError> {
+        if self.state.is_extension() {
+            return Ok(());
+        }
+
```
themighty1, line 215: do we want to do `self.state.replace(State::Error).into_initialized()` at this point?

themighty1 (changes requested):
gw. we need to resolve a potential security issue with reusing a leaky delta.

ot/mpz-ot/src/kos/receiver.rs
sinui0: I'm not sure I follow. This doesn't reuse anything, it simply returns if the Receiver is already set up. The core layer does not allow re-extension

ot/mpz-ot/src/kos/receiver.rs
sinui0: This sets up the base OT protocol if needed, but does not _execute_ the base OTs.

sinui0: This is where Base OTs are actually performed. This implementation does not allow delta to be reused.

https://github.com/privacy-scaling-explorations/mpz/blob/0e0e28b58b3596c2d5551c19077e24099d0325f0/ot/mpz-ot/src/kos/receiver.rs#L189-L198

ot/mpz-ot/src/kos/sender.rs
sinui0: Distinguish between "setting up" the base OT and "performing" the base OT. Some protocols require some sort of 1 time setup. In the case of CO15, the Sender sends his public key to the Receiver.

th4s (approved):
nice, ack

ot/mpz-ot/src/kos/receiver.rs
themighty1: I think what I described above still applies in the following scenario:

- `S` calls `sender.setup()` and then calls `sender.extend()` which performs the correlation check which leaks some bits of delta to `R`
- After some time `S` wants to perform a new OT extension session, so he re-uses his `sender` object by calling `sender.setup()`. At this point no setup actually takes place but the previous setup is re-used (which has a leaky delta)
- `S` proceeds to call `sender.extend()` with a leaky delta

does this sound correct?

sinui0: > S proceeds to call sender.extend() with a leaky delta

Returns an error because KOS can not be extended multiple times. This is guarded at the core layer.

https://github.com/privacy-scaling-explorations/mpz/blob/60ba0131c1f39618e8bcbed2b5797bfaa993c3df/ot/mpz-ot-core/src/kos/sender.rs#L116-L121

themighty1: ok, thanks, makes sense.

themighty1, line 158: In what situations will it make sense for KOS to re-use baseOT delta?  I cant think of any. Maybe we should only allow calling `setup()` in `state.is_initialized()` ?

ot/mpz-ot/src/kos/sender.rs
```diff
@@ -233,6 +196,64 @@ where
...
+    async fn setup<
+        Si: IoSink<Message<BaseOT::Msg>> + Send + Unpin,
+        St: IoStream<Message<BaseOT::Msg>> + Send + Unpin,
+    >(
+        &mut self,
+        sink: &mut Si,
+        stream: &mut St,
+    ) -> Result<(), OTError> {
+        if self.state.is_extension() {
+            return Ok(());
```
themighty1, line 213: In what situations will it make sense for KOS to re-use baseOT delta? I cant think of any. Maybe we should only allow calling setup() in state.is_initialized() ?

sinui0: We never re-use delta, can you explain why you think this is possible?

The function simply returns if the Sender is already set up, that does not mean delta gets reused. It is impossible to reuse delta, that is enforced by the type system in the core layer and the guard in `extend`

themighty1: I was trying to point out that there seems to be no situation where calling `setup` twice makes sense.
If the user calls `setup` twice, it is a sign of the user making an error.
That's why I was suggesting to put there:
```rust
if ! self.state.is_initialized() {
            return Err(...);
        }
```

sinui0: > there seems to be no situation where calling setup twice makes sense.

The reason I took this approach is because it is foreseeable that an already set up instance of a protocol will be injected as a dependency into another. The dependent protocol shouldn't need to be aware of the inner state. If calling setup twice was actually an error or insecure, then we would return an error. But I don't see the issue with it

themighty1: Ok, I see the reason.
This approach makes sense in CO15 because there we have a dedicated `Setup` state. But here the absence of the `Setup` state complicates things.
I will have a closer look how easy it would be to add the `Setup` state.

themighty1: adding a `Setup` state doesn't seem to be trivial. if I feel motivated enough, i'll open a new PR for it.

themighty1 (approved):
gw, ack

ot/mpz-ot/src/kos/sender.rs
sinui0: yes, that makes sense I think. I switched it around so they perform cointoss before the base OT setup.

themighty1: @sinui0  , I dont think your last commit addressed this.
in `receiver.rs` as soon as we enter `setup` we set the error state:
```rust
let ext_receiver = self
            .state
            .replace(State::Error)
            .into_initialized()
            .map_err(ReceiverError::from)?;
```

here we should similarly set an error state immediately upon entering `setup` in case if an error is returned during cointoss or base OT setup.

sinui0: I didn't mean to ignore this today, it slipped my mind. Opened an issue to remember #55

PR #54, opened 2023-08-31
Add LPN and GGM related

xiangxiecrypto:
Please do not merge.

This PR mainly adds two functionalities.
1. Computation of LPN (Learning Parity with Noise) function and its optimization with multi threads.
2. GGM reconstruction.


xiangxiecrypto:
Tried different ways to optimize LPN computation with multi threads.

The problem: Given two F_{2^128} vectors `x`, `e` and a fixed sparse binary matrix `A`, compute `y = Ax + e`.  Each row of `A` has at most `D` non-zero elements.

Since the dimensions of `A` are quite large, `A` is determined by a seed `seed` and is generated on-the-fly.

Optimizations:
1. Single batch computation: compute a batch of rows in one operation.
     a. batch_size = 4.
     b. batch_size = 8.

2. Multi threads: use Rayon.
    a. use `par_chunks_exact_mut` to split the rows of `A` into chunks with length `4` or `8` (depends on 1 above).
    b. use customized threads, and split the rows of `A` into `#rows/#threads` chunks.

My experiments show that 1.a (1.b) + 2.a outperform others.

sinui0:
Is this about ready for review?

xiangxiecrypto:
> Is this about ready for review?

Not yet, I am optimizing ggm, probably will finish in these days.

xiangxiecrypto:
The computation of generation (and reconstruction) of ggm tree in each level is highly parallel. Try to optimize this part in two ways.
1. Use multi threads to compute each level. This turns out to be a bad idea. The cost of thread switching is high.
2. Focus on the last level. This is because the overhead to compute the last level is about 50% of the overall overhead (it is a binary tree). However, although I use multi threads to compute the last level, the improvement is tiny (sometimes even worse). The possible reason is that the underlying `tkprp` is already very fast, it is optimized and paralleled by the complier.

So, I will keep the original implementation, and add comments on public functions.

xiangxiecrypto:
It is ready for review. :)

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+/// Note that in the standard LPN problem, `x` is a binary vector, `e` is a sparse binary vector. The way we difined here is a more generic way in term of computing `y`.
+pub struct Lpn<const D: usize> {
+    // The seed to generate the random sparse matrix A.
+    seed: Block,
+
+    // The length of the secret, i.e., x.
+    k: u32,
+
+    // A mask to optimize reduction operation.
+    mask: u32,
```
sinui0: We can use doc comments for these and they will be shown with `cargo doc --document-private-items`

```suggestion
    /// The seed to generate the random sparse matrix A.
    seed: Block,

    /// The length of the secret, i.e., x.
    k: u32,

    /// A mask to optimize reduction operation.
    mask: u32,
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use crate::{prp::Prp, Block};
+use rayon::prelude::*;
+/// A struct related to LPN.
+/// The `seed` defines a sparse binary matrix `A` with at most `D` non-zero values in each row.
+/// `A` - is a binary matrix with `k` columns and `n` rows. The concrete number of `n` is determined by the input length. `A` will be generated on-the-fly.
+/// `x` - is a `F_{2^128}` vector with length `k`.
+/// `e` - is a `F_{2^128}` vector with length `n`.
+/// Given a vector `x` and `e`, compute `y = Ax + e`.
+/// Note that in the standard LPN problem, `x` is a binary vector, `e` is a sparse binary vector. The way we difined here is a more generic way in term of computing `y`.
```
sinui0: rustdocs need an extra newline

```suggestion

/// A struct related to LPN.
///
/// The `seed` defines a sparse binary matrix `A` with at most `D` non-zero values in each row.
///
/// Given a vector `x` and `e`, compute `y = Ax + e`, where:
///
/// * `A` - is a binary matrix with `k` columns and `n` rows. The concrete number of `n` is determined by the input length. `A` will be generated on-the-fly.
/// * `x` - is a `F_{2^128}` vector with length `k`.
/// * `e` - is a `F_{2^128}` vector with length `n`.
///
/// Note that in the standard LPN problem, `x` is a binary vector, `e` is a sparse binary vector. The way we defined here is a more generic way in term of computing `y`.
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
+//! Implement LPN with local linear code.
+//! More especifically, a local linear code is a random boolean matrix with at most D non-zero values in each row.
+
+use crate::{prp::Prp, Block};
+use rayon::prelude::*;
+/// A struct related to LPN.
```
sinui0: Is there a better description of this? An "LPN encoder" perhaps?

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    // The length of the secret, i.e., x.
+    k: u32,
+
+    // A mask to optimize reduction operation.
+    mask: u32,
+}
+
+impl<const D: usize> Lpn<D> {
+    /// New an LPN instance
```
sinui0:
```suggestion
    /// Creates a new LPN instance.
    ///
    /// # Arguments
    ///
    /// * `seed` - The seed to generate the random sparse matrix A.
    /// * `k` - The length of the secret, i.e., `|x|`.
```

mpz-core/src/lpn.rs
sinui0, line 1: Can we encapsulate the "naive" implementation into a different struct, eg `LpnNaive` and move it into the `mod test` module at the bottom? This will ensure it does not get compiled outside of test config.

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+            for ind in index[i * D..(i + 1) * D].iter_mut() {
+                *ind &= self.mask;
+                *ind = if *ind >= self.k { *ind - self.k } else { *ind };
+
+                *y ^= x[*ind as usize];
+            }
+        }
+    }
+
+    // Compute 4 rows as a batch, this is for the `compute` function.
```
sinui0:
```suggestion
    /// Compute 4 rows as a batch, this is for the `compute` function.
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+            for ind in index[i * D..(i + 1) * D].iter_mut() {
+                *ind &= self.mask;
+                *ind = if *ind >= self.k { *ind - self.k } else { *ind };
+
+                *y ^= x[*ind as usize];
+            }
+        }
+    }
+
+    // Compute one row.
```
sinui0: This comment doesn't provide any more information than the function signature

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    // Compute 4 rows as a batch, this is for the `compute` function.
+    #[inline]
+    fn compute_four_rows_indep(&self, y: &mut [Block], x: &[Block], pos: usize, prp: &Prp) {
+        let mut cnt = 0u64;
+        let index = [0; D].map(|_| {
+            let i = cnt;
+            cnt += 1;
+            Block::from(bytemuck::cast::<_, [u8; 16]>([pos as u64, i]))
+        });
```
sinui0: `std::array::from_fn` is handy here:

```suggestion
        let index: [Block; D] = std::array::from_fn(|_| {
            let i = cnt;
            cnt += 1;
            Block::from(bytemuck::cast::<_, [u8; 16]>([pos as u64, i]))
        });
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /// Compute `Ax + e` with multiple threads.
+    /// Input: `x` with length `k`.
+    /// Input: `y` with length `n`, this is actually `e` in LPN.
+    /// Output: `y = Ax + y`.
+    pub fn compute(&self, y: &mut [Block], x: &[Block]) {
+        assert_eq!(x.len() as u32, self.k);
+        assert!(x.len() >= D);
+        let prp = Prp::new(self.seed);
```
sinui0, line 93: `self.seed` is immutable, we should add it as a field to `Lpn` and instantiate it in the constructor

mpz-core/benches/lpn.rs
```diff
@@ -0,0 +1,103 @@
...
+        let k = 5_060;
+        let n = 166_400;
+        let lpn = Lpn::<10>::new(seed, k);
+        let mut x = vec![Block::ZERO; k as usize];
+        let mut y = vec![Block::ZERO; n];
+        let mut prg = Prg::new();
+        prg.random_blocks(&mut x);
+        prg.random_blocks(&mut y);
+        bench.iter(|| {
+            black_box(lpn.compute_naive(&mut y, &x));
```
sinui0: > passing a unit value to a function
> for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg

mpz-core/Cargo.toml
```diff
@@ -24,7 +24,8 @@ itybity.workspace = true
 opaque-debug.workspace = true
 bcs = "0.1.5"
 rand_core = "0.6.4"
-bytemuck = {workspace = true, features = ["derive"]}
+bytemuck = { workspace = true, features = ["derive"] }
+rayon = { version = "1.7.0" }
```
sinui0: Please put this under a feature flag:

```toml
[features]
default = ["cointoss", "rayon"]
cointoss = ["dep:rand_chacha"]
rayon = ["dep:rayon"]

[dependencies]
rayon = { workspace = true, optional = true }
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+        let size = y.len() - (y.len() % 4);
+
+        y.par_chunks_exact_mut(4).enumerate().for_each(|(i, y)| {
+            self.compute_four_rows_indep(y, x, i * 4, &prp);
+        });
```
sinui0:
```toml
[dependencies]
cfg-if.workspace = true
```

```suggestion
        cfg_if::cfg_if! {
            if #[cfg(feature = "rayon")] {
                let iter = y.par_chunks_exact_mut(4).enumerate();
            } else {
                let iter = y.chunks_exact_mut(4).enumerate();
            }
        }

        iter.for_each(|(i, y)| {
            self.compute_four_rows_indep(y, x, i * 4, &prp);
        });
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        for i in batch_size * 4..y.len() {
+            self.compute_one_row(y, x, i, &prp);
+        }
+    }
+
+    /// Compute `Ax + e` with multiple threads.
+    /// Input: `x` with length `k`.
+    /// Input: `y` with length `n`, this is actually `e` in LPN.
+    /// Output: `y = Ax + y`.
```
sinui0:
```suggestion
    /// Computes `Ax + e`, writing the result in-place into `y`.
    ///
    /// # Arguments
    ///
    /// * `x` - Secret vector with length `k`.
    /// * `y` - Error vector with length `n`. This is actually `e` in LPN.
    ///
    /// # Panics
    ///
    /// Panics if `x.len() != k` or `y.len() != n`.
```

mpz-core/src/prp.rs
```diff
@@ -0,0 +1,32 @@
+//! An implementation of Pseudo Random Permutation (PRP) based on AES.
+
+use crate::{aes::AesEncryptor, Block};
+
+/// Struct of PRP
```
sinui0:
```suggestion
/// Pseudo Random Permutation (PRP) based on AES
```

mpz-core/src/prp.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// Permute many blocks.
+    #[inline(always)]
+    pub fn permute_many_blocks<const N: usize>(&self, blks: [Block; N]) -> [Block; N] {
+        self.0.encrypt_many_blocks(blks)
+    }
+
+    /// Permute block slice.
+    #[inline(always)]
+    pub fn permute_block_slice(&self, blks: &mut [Block]) {
```
sinui0:
```suggestion
    pub fn permute_blocks_inplace(&self, blks: &mut [Block]) {
```

mpz-core/src/aes.rs
```diff
@@ -99,6 +100,24 @@ impl AesEncryptor {
...
+            );
+            blks[i * AesEncryptor::AES_BLOCK_COUNT..(i + 1) * AesEncryptor::AES_BLOCK_COUNT]
+                .copy_from_slice(&self.encrypt_many_blocks(buf));
+        }
+
+        let remain = len % AesEncryptor::AES_BLOCK_COUNT;
+        for block in blks[len - remain..].iter_mut() {
+            *block = self.encrypt_block(*block);
+        }
+    }
```
sinui0: See #59

mpz-core/src/ggm_tree.rs
```diff
@@ -18,15 +18,15 @@ impl GgmTree {
...
+    /// Output: `k0` - XORs of all the left-node values in each level, with size `depth`.
+    /// Output: `k1`- XORs of all the right-node values in each level, with size `depth`.
     /// This implementation is adapted from EMP Toolkit.
     pub fn gen(&self, seed: Block, tree: &mut [Block], k0: &mut [Block], k1: &mut [Block]) {
-        assert!(tree.len() == 1 << (self.depth - 1));
-        assert!(k0.len() == self.depth - 1);
-        assert!(k1.len() == self.depth - 1);
+        assert!(tree.len() == 1 << (self.depth));
+        assert!(k0.len() == self.depth);
+        assert!(k1.len() == self.depth);
```
sinui0:
```suggestion
        assert_eq!(tree.len(), 1 << (self.depth));
        assert_eq!(k0.len(), self.depth);
        assert_eq!(k1.len(), self.depth);
```

mpz-core/src/ggm_tree.rs
```diff
@@ -18,15 +18,15 @@ impl GgmTree {
...
         let mut buf = vec![Block::ZERO; 8];
```
sinui0: I think we can avoid this heap allocation

```suggestion
        let mut buf = [Block::ZERO; 8];
```

mpz-core/src/ggm_tree.rs
```diff
@@ -18,15 +18,15 @@ impl GgmTree {
...
-    /// Input: `seed`: a seed.
-    /// Output: `tree`: a GGM (binary tree) `tree`, with size `2^{depth-1}`
-    /// Output: `k0`: XORs of all the left-node values in each level, with size `depth-1`.
-    /// Output: `k1`: XORs of all the right-node values in each level, with size `depth-1`.
+    /// Input : `seed` - a seed.
+    /// Output: `tree` - a GGM (binary tree) `tree`, with size `2^{depth}`.
+    /// Output: `k0` - XORs of all the left-node values in each level, with size `depth`.
+    /// Output: `k1`- XORs of all the right-node values in each level, with size `depth`.
     /// This implementation is adapted from EMP Toolkit.
     pub fn gen(&self, seed: Block, tree: &mut [Block], k0: &mut [Block], k1: &mut [Block]) {
```
sinui0:
```suggestion
    /// Generate a GGM tree in-place.
    /// 
    /// # Arguments
    /// 
    /// * `seed` - a seed.
    /// * `tree` - the destination to write the GGM (binary tree) `tree`, with size `2^{depth}`.
    /// * `k0` - XORs of all the left-node values in each level, with size `depth`.
    /// * `k1`- XORs of all the right-node values in each level, with size `depth`.
    pub fn gen(&self, seed: Block, tree: &mut [Block], k0: &mut [Block], k1: &mut [Block]) {
      // This implementation is adapted from EMP Toolkit.
```

mpz-core/src/ggm_tree.rs
```diff
@@ -56,43 +56,97 @@ impl GgmTree {
             }
         }
     }
+
+    /// Reconstruct the GGM tree except the value in a given position.
+    /// Input : `k` - a slice of blocks with length `depth`, the values of k are chosen via OT from k0 and k1. For the i-th value, if alpha[i] == 1, k[i] = k1[i]; else k[i] = k0[i].
+    /// Input : `alpha` - a slice of bits with length `depth`.
+    /// Output : `tree` - the ggm tree, except `tree[pos] == Block::ZERO`. The bit decomposition of `pos` is the complement of `alpha`. I.e., `pos[i] = 1 xor alpha[i]`.
```
sinui0:
```suggestion
    /// Reconstruct the GGM tree except the value in a given position.
    /// 
    /// This reconstructs the GGM tree entirely except `tree[pos] == Block::ZERO`. The bit decomposition of `pos` is the complement of `alpha`. i.e., `pos[i] = 1 xor alpha[i]`.
    /// 
    /// # Arguments
    /// 
    /// * `k` - a slice of blocks with length `depth`, the values of k are chosen via OT from k0 and k1. For the i-th value, if alpha[i] == 1, k[i] = k1[i]; else k[i] = k0[i].
    /// * `alpha` - a slice of bits with length `depth`.
    /// * `tree` - the destination to write the GGM tree.
```

mpz-core/src/ggm_tree.rs
```diff
@@ -56,43 +56,97 @@ impl GgmTree {
...
+        for i in (start..sz).step_by(2) {
+            sum ^= tree[i];
+        }
+        tree[pos] = sum ^ k;
+
+        if depth == (self.depth) {
+            return;
+        }
+
+        let mut buf = vec![Block::ZERO; 8];
```
sinui0:
```suggestion
        let mut buf = [Block::ZERO; 8];
```

mpz-core/src/prp.rs
```diff
@@ -0,0 +1,32 @@
+//! An implementation of Pseudo Random Permutation (PRP) based on AES.
+
+use crate::{aes::AesEncryptor, Block};
+
+/// Struct of PRP
+pub struct Prp(AesEncryptor);
+
+impl Prp {
+    /// New an instance of Prp.
```
sinui0:
```suggestion
    /// Creates a new Prp instance.
```

sinui0 (changes requested):
Looks good, thank you!

I mostly provided some feedback around formatting conventions and idiomatic Rust things. Looks good otherwise

You may want to rebase on #59, see the relevant comment below.

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
...
+///
+/// Given a vector `x` and `e`, compute `y = Ax + e`.
+///
+/// `A` - is a binary matrix with `k` columns and `n` rows. The concrete number of `n` is determined by the input length. `A` will be generated on-the-fly.
+///
+/// `x` - is a `F_{2^128}` vector with length `k`.
+///
+/// `e` - is a `F_{2^128}` vector with length `n`.
+///
+/// Note that in the standard LPN problem, `x` is a binary vector, `e` is a sparse binary vector. The way we difined here is a more generic way in term of computing `y`.
```
themighty1, line 18: @xiangxiecrypto , would it be possible to add some links to literature which justifies mixing a binary matrix with non-binary vectors? I could not find it with a simple web search.

xiangxiecrypto: It is a common step in PCG-style COT. You can find the intuition in the [Ferret paper](https://eprint.iacr.org/2020/924.pdf) (page 11) and the references in it.
Note that we can embed a bit into a `Block` to support binary matrix with binary vectors.

> @xiangxiecrypto , would it be possible to add some links to literature which justifies mixing a binary matrix with non-binary vectors? I could not find it with a simple web search.

themighty1: @xiangxiecrypto , thanks, I was worried because I read in
https://eprint.iacr.org/2017/617
"""
How to sample matrices with large dual distance? We suggest to sample a d-sparse matrix M ∈ Fm×k in two steps. First, choose the locations of the non-zero entries of the matrix (e.g., by selecting a random set of d entries per row), and then fill them with random field elements.
"""
since it says  "random field elements" I was assuming that we can't just use "1" as your code does but we must use an element from F_2^128.

Maybe this does not apply to our case, idk. If it's not too laborious to explain, could you shed some light why using "1" instead of a random field element is ok?

xiangxiecrypto: > @xiangxiecrypto , thanks, I was worried because I read in https://eprint.iacr.org/2017/617 """ How to sample matrices with large dual distance? We suggest to sample a d-sparse matrix M ∈ Fm×k in two steps. First, choose the locations of the non-zero entries of the matrix (e.g., by selecting a random set of d entries per row), and then fill them with random field elements. """ since it says "random field elements" I was assuming that we can't just use "1" as your code does but we must use an element from F_2^128.
> 
> Maybe this does not apply to our case, idk. If it's not too laborious to explain, could you shed some light why using "1" instead of a random field element is ok?

The security here relies on LPN over F_2, therefore you only need to choose bits.
The reason the function also applies to F_2^128 is that the protocol requires these operations (not related to security assumption). See page 11 in the Ferret paper.

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
+//! Implement LPN with local linear code.
+//! More especifically, a local linear code is a random boolean matrix with at most D non-zero values in each row.
+
+use crate::{prp::Prp, Block};
+use rayon::prelude::*;
+/// An LPN encoder.
+///
+/// The `seed` defines a sparse binary matrix `A` with at most `D` non-zero values in each row.
```
themighty1, line 8: @xiangxiecrypto , is **at most** correct here? Shouldn't it be **exactly** D non-zero values?

xiangxiecrypto: It is **at most**. Recall that the process is choosing `d` positions at random in a row, and filling the `d` positions with random elements (in F_2). You do not need to fill all `d` positions with `1`.

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// * `x` - Secret vector with length `k`.
+    /// * `y` - Error vector with length `n`, this is actually `e` in LPN.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `x.len() !=k` or `y.len() != n`.
+    pub fn compute(&self, y: &mut [Block], x: &[Block]) {
+        assert_eq!(x.len() as u32, self.k);
+        assert!(x.len() >= D);
```
themighty1, line 92: @xiangxiecrypto , shouldn't we have here a specific algorithm which bounds the minimal amount of rows/columns depending on D?

As mentioned in https://eprint.iacr.org/2017/617
"Then, for 80-bit security and d = 10 it turns out that we will need approximately k = 182 columns and k 2 rows, while for 100-bit security we need k = 240."

xiangxiecrypto: The security also depends on `k` and Hamming weight of the error vector. I will specify a new struct to contain these parameters (together with others) when I implement the COT protocol. We could simply focus on `D = 10` here.
It is only for computation purpose here, maybe I should change the name to `LpnEncoder`

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `x` - Secret vector with length `k`.
+    /// * `y` - Error vector with length `n`, this is actually `e` in LPN.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `x.len() !=k` or `y.len() != n`.
+    pub fn compute(&self, y: &mut [Block], x: &[Block]) {
```
themighty1, line 90: What are your thoughts about having one common method for computing rows regardless of row count? The code would look like this:

```rust
/// The size of one batch of matrix row computation.
///
/// Experiments show that 4 gives the best performance with rayon.
const BATCH_SIZE: usize = 4;

    pub fn compute(&self, y: &mut [Block], x: &[Block]) {
        assert_eq!(x.len() as u32, self.k);
        assert!(x.len() >= D);
        let prp = Prp::new(self.seed);

        // how many rows will be processed in batches
        let rows_in_batches = y.len() - (y.len() % BATCH_SIZE);

        cfg_if::cfg_if! {
            if #[cfg(feature = "rayon")]{
                let mut iter = y.par_chunks_exact_mut(BATCH_SIZE);
            }else{
                let mut iter = y.par_chunks_exact_mut(BATCH_SIZE);
            }
        }

        let remaining_rows = iter.take_remainder();

        iter.enumerate().for_each(|(i, y)| {
            self.compute_rows(y, x, i * BATCH_SIZE, &prp);
        });

        // process remaining rows, if any
        self.compute_rows(remaining_rows, x, rows_in_batches, &prp);
    }

   /// Computes as many rows as there are elements in `y`, placing the result in `y`.
    #[inline]
    fn compute_rows(&self, y: &mut [Block], x: &[Block], pos: usize, prp: &Prp) {
        // How many `Blocks` needed to derive random u32-sized indices for one row
        let block_cnt = (D + 4 - 1) / 4;

        let mut indices = (0..y.len())
            .flat_map(|offset| {
                (0..block_cnt).map(move |i| {
                    Block::from(bytemuck::cast::<_, [u8; 16]>([
                        (pos + offset) as u64,
                        i as u64,
                    ]))
                })
            })
            .collect::<Vec<Block>>();

        // derive pseudo-random u32-sized indices
        prp.permute_block_inplace(&mut indices);
        let indices = bytemuck::cast_slice_mut::<_, u32>(&mut indices);

        for (row, indices) in y.iter_mut().zip(indices.chunks_exact_mut(block_cnt * 4)) {
            // We only need D indices for one row, ignore any extra ones
            for ind in indices.iter_mut().take(D) {
                // reduce the index to be in the [0, k) range
                *ind &= self.mask;
                *ind = if *ind >= self.k { *ind - self.k } else { *ind };

                *row ^= x[*ind as usize];
            }
        }
    }
```

mpz-core/src/ggm_tree.rs
```diff
@@ -18,16 +18,20 @@ impl GgmTree {
...
-    /// Output: `tree`: a GGM (binary tree) `tree`, with size `2^{depth-1}`
-    /// Output: `k0`: XORs of all the left-node values in each level, with size `depth-1`.
-    /// Output: `k1`: XORs of all the right-node values in each level, with size `depth-1`.
-    /// This implementation is adapted from EMP Toolkit.
+    /// Create a GGM tree in-place.
+    ///
+    /// # Arguments
+    ///
+    /// * `seed` - a seed.
+    /// * `tree` - the destination of write the GGM (binary tree) `tree`, with size `2^{depth}`.
```
themighty1:
```suggestion
    /// * `tree` - the destination to write the GGM (binary tree) `tree`, with size `2^{depth}`.
```

mpz-core/src/ggm_tree.rs
```diff
@@ -37,7 +41,7 @@ impl GgmTree {
         k1[1] = buf[1] ^ buf[3];
         tree[0..4].copy_from_slice(&buf[0..4]);
 
-        for h in 2..self.depth - 1 {
+        for h in 2..self.depth {
             k0[h] = Block::ZERO;
             k1[h] = Block::ZERO;
             let sz = 1 << h;
```
themighty1, line 49:
```suggestion

            // How many nodes there are in this layer
            let sz = 1 << h;
```

mpz-core/src/ggm_tree.rs
```diff
@@ -56,43 +60,102 @@ impl GgmTree {
...
+    // Handle each layer.
+    fn reconstruct_layer(
+        &self,
+        depth: usize,
+        left_or_right: bool,
+        pos: usize,
+        k: Block,
+        tree: &mut [Block],
+    ) {
+        let sz = 1 << depth;
```
themighty1, line 104:
```suggestion
        // How many nodes there are in this layer
        let sz = 1 << depth;
```

themighty1 (changes requested):
gw, thank you.
I have not fully reviewed `reconstruct_layer`. will do some time next week.

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
...
+            self.compute_four_rows_indep(y, x, i * 4, &prp);
+        });
+
+        for i in size..y.len() {
+            self.compute_one_row(y, x, i, &prp);
+        }
+    }
+}
+
+mod tests {
```
sinui0, line 115:
```suggestion
#[cfg(test)]
mod tests {
```

mpz-core/src/lpn.rs
```diff
@@ -0,0 +1,180 @@
...
+        assert_eq!(x.len() as u32, self.k);
+        assert!(x.len() >= D);
+        let prp = Prp::new(self.seed);
+        let size = y.len() - (y.len() % 4);
+
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "rayon")]{
+                let iter = y.par_chunks_exact_mut(4).enumerate();
+            }else{
+                let iter = y.par_chunks_exact_mut(4).enumerate();
```
sinui0:
```suggestion
                let iter = y.chunks_exact_mut(4).enumerate();
```

mpz-core/src/ggm_tree.rs
```diff
@@ -56,43 +60,102 @@ impl GgmTree {
...
+    ///
+    /// This reconstructs the GGM tree entirely except `tree[pos] == Block::ZERO`. The bit decomposition of `pos` is the complement of `alpha`. i.e., `pos[i] = 1 xor alpha[i]`.
+    ///
+    /// # Arguments
+    ///
+    /// * `k` - a slice of blocks with length `depth`, the values of k are chosen via OT from k0 and k1. For the i-th value, if alpha[i] == 1, k[i] = k1[i]; else k[i] = k0[i].
+    /// * `alpha` - a slice of bits with length `depth`.
+    /// * `tree` - the destination to write the GGM tree.
+    pub fn reconstruct(&self, tree: &mut [Block], k: &[Block], alpha: &[bool]) {
+        let mut pos = 0;
```
themighty1, line 80: do we want some asserts on `tree.len()`, `k.len()`, `alpha.len()` similarly like like in `gen()` above ?

mpz-core/src/lpn.rs
xiangxiecrypto: good point, let me check

mpz-core/src/ggm_tree.rs
xiangxiecrypto: yes, we need that asserts.

mpz-core/src/lpn.rs
xiangxiecrypto: It seems that the test failed.

themighty1: Oh, I see.
Looking at this again:
earlier in `compute_four_rows_indep` there were exactly 10 PRG calls for 4 rows, i.e. 2.5 calls/row
This new code makes 3 calls/row.
This is 20% overhead which is too much.

Makes sense to continue with the `compute_four_rows_indep` approach.

mpz-core/src/aes.rs
sinui0: @xiangxiecrypto you should be able to rebase on dev and use the existing impl for this now

sinui0 (commented):
Almost there!

xiangxiecrypto:
Miri CI tests fail on `Clmul`, would you please help to fix it? @sinui0 @themighty1

sinui0:
> Miri CI tests fail on `Clmul`, would you please help to fix it? @sinui0 @themighty1

#74 should be merged soon

sinui0:
You can rebase now, CI should be fixed

sinui0 (approved):
:rocket:

themighty1 (approved):

PR #59, opened 2023-09-11
feat: generic array casts

sinui0:
This PR adds helper methods for converting variations of `Block` into `GenericArray<u8, U16>` from the `generic-array` crate. This reduces copies in a few places and lets us take better advantage of instruction-level parallelism provided by RustCrypto crates.

### DIFF
--- a/mpz-core/src/aes.rs
+++ b/mpz-core/src/aes.rs
@@ -1,7 +1,7 @@
 //! Fixed-key AES cipher
 
-use aes::{cipher::generic_array::GenericArray, Aes128, Aes128Enc};
-use cipher::{generic_array::functional::FunctionalSequence, BlockEncrypt, KeyInit};
+use aes::Aes128Enc;
+use cipher::{BlockEncrypt, KeyInit};
 use once_cell::sync::Lazy;
 
 use crate::Block;
@@ -13,12 +13,12 @@ pub const FIXED_KEY: [u8; 16] = [
 
 /// Fixed-key AES cipher
 pub static FIXED_KEY_AES: Lazy<FixedKeyAes> = Lazy::new(|| FixedKeyAes {
-    aes: Aes128::new_from_slice(&FIXED_KEY).unwrap(),
+    aes: Aes128Enc::new_from_slice(&FIXED_KEY).unwrap(),
 });
 
 /// Fixed-key AES cipher
 pub struct FixedKeyAes {
-    aes: Aes128,
+    aes: Aes128Enc,
 }
 
 impl FixedKeyAes {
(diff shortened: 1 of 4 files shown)


mpz-core/src/block.rs
```diff
@@ -121,6 +121,40 @@ impl Block {
...
+
+    /// Converts a slice of blocks to a slice of [`GenericArray<u8, U16>`](cipher::generic_array::GenericArray)
+    /// from the [`generic-array`](https://docs.rs/generic-array/latest/generic_array/) crate.
+    #[allow(dead_code)]
+    pub(crate) fn as_generic_array_slice(slice: &[Self]) -> &[GenericArray<u8, U16>] {
+        // # Safety
+        // This is always safe because `Block` and `GenericArray<u8, U16>` have the same memory layout.
+        // See https://github.com/fizyk20/generic-array/blob/37dc6aefc3ed5c423ad7402d4febf06a3e78a223/src/lib.rs#L838-L845
+        // TODO: Use methods provided by `generic-array` once 1.0 is released.
+        unsafe { std::mem::transmute(slice) }
```
th4s, line 146: We rely here on an implementation detail of an external library. What happens if the memory layout of `GenericArray` is changed upstream?

mpz-core/src/block.rs
```diff
@@ -121,6 +121,40 @@ impl Block {
...
+    }
+
+    /// Converts a mutable slice of blocks to a mutable slice of [`GenericArray<u8, U16>`](cipher::generic_array::GenericArray)
+    /// from the [`generic-array`](https://docs.rs/generic-array/latest/generic_array/) crate.
+    pub(crate) fn as_generic_array_mut_slice(slice: &mut [Self]) -> &mut [GenericArray<u8, U16>] {
+        // # Safety
+        // This is always safe because `Block` and `GenericArray<u8, U16>` have the same memory layout.
+        // See https://github.com/fizyk20/generic-array/blob/37dc6aefc3ed5c423ad7402d4febf06a3e78a223/src/lib.rs#L847-L854
+        // TODO: Use methods provided by `generic-array` once 1.0 is released.
+        unsafe { std::mem::transmute(slice) }
```
th4s, line 156: same here

th4s (commented):
:+1:

mpz-core/src/block.rs
sinui0: `GenericArray` is `#[repr(transparent)]`, so it will always have the same memory layout as `[u8; N]`. I generally agree though, its not the best, so we should remove this as soon as `1.0` is accepted into the `cipher` crate. For now, our workspace is pinned to `0.14`, and assuming they adhere to `semver` then this is fine.

themighty1: @sinui0 , could you pls add a note to Cargo.toml explaining why `generic-array` should stay at 0.14

mpz-core/src/aes.rs
```diff
@@ -30,40 +30,96 @@ impl FixedKeyAes {
...
-        Block::from(x) ^ block
+        let mut h = block;
+        self.aes.encrypt_block(h.as_generic_array_mut());
+        h ^ block
+    }
+
+    /// Correlation-robust hash function instantiated using fixed-key AES
+    /// (cf. <https://eprint.iacr.org/2019/074>, §7.2).
+    ///
+    /// `π(x) ⊕ x`, where `π` is instantiated using fixed-key AES.
```
themighty1, line 87:
```suggestion
    /// `π(x) ⊕ x`, where `π` is instantiated using fixed-key AES.
    /// 
    /// # Arguments
    ///
    /// * `blocks` - The blocks to hash in-place.
```

mpz-core/src/aes.rs
```diff
@@ -30,40 +30,96 @@ impl FixedKeyAes {
...
     pub fn ccr(&self, block: Block) -> Block {
         self.cr(Block::sigma(block))
     }
+
+    /// Circular correlation-robust hash function instantiated using fixed-key AES
+    /// (cf.<https://eprint.iacr.org/2019/074>, §7.3).
+    ///
+    /// `π(σ(x)) ⊕ σ(x)`, where `π` is instantiated using fixed-key AES
+    ///
+    /// See [`Block::sigma`](Block::sigma) for more details on `σ`.
```
themighty1, line 121:
```suggestion
    /// See [`Block::sigma`](Block::sigma) for more details on `σ`.
    ///
    /// # Arguments
    ///
    /// * `blocks` - The blocks to hash in-place.
```

mpz-core/src/aes.rs
```diff
@@ -83,34 +139,49 @@ impl AesEncryptor {
...
+        self.0
+            .encrypt_blocks(Block::as_generic_array_mut_slice(blks));
     }
 
     /// Encrypt many blocks with many keys.
-    /// Input: `NK` AES keys `keys`, and `NK * NM` blocks `blks`
-    /// Output: each batch of NM blocks encrypted by a corresponding AES key.
-    /// Only handle the first NK * NM blocks of blks, do not handle the rest.
+    ///
+    /// Each batch of NM blocks are encrypted by a corresponding AES key.
```
themighty1:
```suggestion
    /// Each batch of NM blocks is encrypted by a corresponding AES key.
```

themighty1 (approved):
ack, some code comments

PR #73, opened 2023-09-28
Refactor `clmul` to use Union

### LINKED ISSUE #44: Refactor out dispatch from `clmul`
Right now the clmul crate has to `match` on whether instrinsics are available for *every operation*, incurring significant overhead.

https://github.com/privacy-scaling-explorations/mpz/blob/a2007e628e1bc52722a3286b1ba053f5d55e2b7e/clmul/src/backend.rs#L86-L124

We should be able to factor this out using the [original approach from the polyval crate](https://github.com/RustCrypto/universal-hashes/blob/master/polyval/src/backend/autodetect.rs) with [`Union`](https://doc.rust-lang.org/reference/items/unions.html)

000wan:
# Summary
It resolves #44.
Struct `Clmul` is modified to:
```rust
pub struct Clmul {
    inner: Inner,
    token: mul_intrinsics::InitToken,
}
```
with Union type `Inner`:
```rust
union Inner {
    intrinsics: intrinsics::Clmul,
    soft: soft::Clmul,
}
```
It follows the implementation of [polyval](https://github.com/RustCrypto/universal-hashes/blob/master/polyval/src/backend/autodetect.rs) using [`InitToken`](https://github.com/RustCrypto/utils/blob/master/cpufeatures/src/lib.rs).

# Results
https://github.com/privacy-scaling-explorations/mpz/pull/73#issuecomment-1741284297

# References
https://doc.rust-lang.org/reference/items/unions.html
https://github.com/RustCrypto/universal-hashes/blob/master/polyval/src/backend/autodetect.rs
https://github.com/RustCrypto/utils/blob/master/cpufeatures/src/lib.rs

### DIFF
--- a/clmul/src/backend.rs
+++ b/clmul/src/backend.rs
@@ -74,6 +74,7 @@ lazy_static! {
 }
 
 #[derive(Clone, Copy)]
+/// Carryless multiplication
 pub struct Clmul {
     inner: Inner,
 }
@@ -95,35 +96,6 @@ impl core::fmt::Debug for Clmul {
         }
     }
 }
-
-/*
-cfg_if! {
-    if #[cfg(any(all(target_arch = "aarch64", feature = "armv8"), any(target_arch = "x86_64", target_arch = "x86")))]{
-        #[derive(Clone, Copy, Debug)]
-        /// Carryless multiplication
-        pub struct Clmul {
-            intrinsics: Option<intrinsics::Clmul>,
-            soft: Option<soft::Clmul>,
-        }
-    } else {
-        #[derive(Clone, Copy, Debug)]
-        /// Carryless multiplication
-        pub struct Clmul {
-            // intrinsics will never be used on a non-supported arch but Rust
-            // won't allow to declare it with a None type, so we need to
-            // provide some type
-            intrinsics: Option<soft::Clmul>,
-            soft: Option<soft::Clmul>,
-        }
-    }
-}
-*/
-
-// #[derive(Clone, Copy)]
-// pub struct Clmul {
-//     intrinsics: Option<intrinsics::Clmul>,
-//     soft: Option<soft::Clmul>,
-// }
 
 impl Clmul {
     pub fn new(h: &[u8; 16]) -> Self {
--- a/clmul/src/backend.rs
+++ b/clmul/src/backend.rs
@@ -56,27 +56,11 @@ cfg_if! {
     }
 }
 
-use lazy_static::lazy_static;
-lazy_static! {
-    static ref IS_INTR: bool = {
-        cfg_if! {
-            if #[cfg(feature = "force-soft")] {
-                false
-            } else if #[cfg(any(all(target_arch = "aarch64", feature = "armv8"), any(target_arch = "x86_64", target_arch = "x86")))]{
-                mul_intrinsics::get()
-            } else {
-                // "force-soft" feature was not enabled but neither was
-                //  supported arch found. Falling back to soft backend.
-                false
-            }
-        }
-    };
-}
-
 #[derive(Clone, Copy)]
 /// Carryless multiplication
 pub struct Clmul {
     inner: Inner,
+    token: mul_intrinsics::InitToken,
 }
 
 #[derive(Clone, Copy)]
@@ -88,7 +72,7 @@ union Inner {
 impl core::fmt::Debug for Clmul {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         unsafe {
-            if *IS_INTR {
+            if self.token.get() {
                 self.inner.intrinsics.fmt(f)
             } else {
                 self.inner.soft.fmt(f)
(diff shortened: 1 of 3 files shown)


sinui0:
Thanks for the PR!

Would it be possible to remove the `match` blocks all-together? The `cpufeatures` macro generates a ZST `InitToken` which can be used to replace `is_intr` in `Clmul`.

https://github.com/RustCrypto/universal-hashes/blob/fc463be29521177527377f02ada141c1ec329ae9/polyval/src/backend/autodetect.rs#L25-L56

This essentially moves the `is_intr` to a global atomic instead of storing a bool in every instance of `Clmul`. We can also avoid checking both instances in this case.

000wan:
Oh, I see. Do you mean declaring `is_intr` for global state using [Atomic module](https://doc.rust-lang.org/std/sync/atomic/)?

clmul/src/backend.rs
```diff
@@ -56,103 +56,94 @@ cfg_if! {
...
 }
 
-cfg_if! {
-    if #[cfg(any(all(target_arch = "aarch64", feature = "armv8"), any(target_arch = "x86_64", target_arch = "x86")))]{
-        #[derive(Clone, Copy, Debug)]
-        /// Carryless multiplication
-        pub struct Clmul {
-            intrinsics: Option<intrinsics::Clmul>,
-            soft: Option<soft::Clmul>,
+use lazy_static::lazy_static;
```
000wan: I used the `lazy_static` crate for the global variable.

000wan:
## [Updated after 570d042]
So I've benched these both for `intrinsics` and `soft`, and here are the results:
(Compared to the `dev` branch, running `$ cargo bench --bench clmul`)

## intrinsics:
```
clmul                   time:   [2.0842 ns 2.0942 ns 2.1036 ns]                   
                        change: [-80.054% -79.865% -79.674%] (p = 0.00 < 0.05)
                        Performance has improved.

clmul_reuse             time:   [6.7801 ns 6.8201 ns 6.8617 ns]                         
                        change: [-18.986% -14.868% -11.035%] (p = 0.00 < 0.05)
                        Performance has improved.

reduce                  time:   [668.48 ps 686.49 ps 705.57 ps]                    
                        change: [-94.706% -94.544% -94.380%] (p = 0.00 < 0.05)
```

## soft:
```
clmul                   time:   [34.806 ns 35.014 ns 35.233 ns]                   
                        change: [-10.990% -9.9367% -8.8572%] (p = 0.00 < 0.05)
                        Performance has improved.

clmul_reuse             time:   [34.388 ns 34.735 ns 35.132 ns]                         
                        change: [-18.041% -16.826% -15.262%] (p = 0.00 < 0.05)
                        Performance has improved.

reduce                  time:   [1.7560 ns 1.7783 ns 1.8069 ns]                    
                        change: [-83.553% -83.150% -82.735%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It seems `clmul` using intrinsics has been considerably improved, and using soft has a little.

sinui0:
Nice results!

I think we can avoid depending on `lazy_static`, we already use `cpufeatures`.

You can do:

```rust
pub struct Clmul {
  inner: Inner,
  token: mul_instrinsics::InitToken
}

impl Clmul {
  pub fn clmul(self, other: Self) -> (Self, Self) {
    if self.token.get() {
      // self.inner is intrinsics
    } else {
      // self.inner is soft
    }
  }
}
```

`InitToken` is a ZST so this should be effectively the same, but 1 less dep

clmul/src/backend.rs
```diff
@@ -56,103 +56,94 @@ cfg_if! {
...
+        cfg_if! {
+            if #[cfg(feature = "force-soft")] {
+                false
+            } else if #[cfg(any(all(target_arch = "aarch64", feature = "armv8"), any(target_arch = "x86_64", target_arch = "x86")))]{
+                mul_intrinsics::get()
+            } else {
+                // "force-soft" feature was not enabled but neither was
+                //  supported arch found. Falling back to soft backend.
+                false
+            }
```
000wan: I also tried to follow the polyval implementation but was slightly concerned that if we use `token` then it becomes difficult to deal with the feature `force-soft`. Shall I just use `token`? @sinui0

sinui0: I think its worth it to avoid another dependency. Which part is complicated? It should only need to be handled in the constructor

000wan: Alright I updated in c0c846d you can check out

clmul/src/backend.rs
```diff
@@ -56,103 +56,79 @@ cfg_if! {
...
+            Inner {
+                soft: soft::Clmul::new(h),
+            }
+        };
+
+        Self { inner, token }
     }
 
     /// Performs carryless multiplication
     pub fn clmul(self, x: Self) -> (Self, Self) {
```
sinui0, line 114:
```suggestion
     #[inline]
     pub fn clmul(self, x: Self) -> (Self, Self) {
```

sinui0 (approved):
Good work!

If you have a moment maybe you could check if inlining helps perf too

Thanks!

clmul/src/backend.rs
000wan: I added and it seems inline only helps for `reduce_gcm` function. I updated the result on https://github.com/privacy-scaling-explorations/mpz/pull/73#issuecomment-1741284297.

themighty1 (approved):
Thank you, looks good!!!
Do you plan to add anything else, or can I merge this @000wan ?

000wan:
I think we can merge this.
Thank you for kind reviews !! :)

PR #74, opened 2023-10-05
fix: clmul aarch64

sinui0:
This PR reorganizes some things and brings the `clmul` crate more inline with the original `polyval` crate.

# Changes
- Removes `armv8` and `force-soft` features in favor of using `RUSTFLAGS`
- Reorganizes modules to match `polyval` layout

### DIFF
--- a/clmul/Cargo.toml
+++ b/clmul/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 
 [dependencies]
 cfg-if.workspace = true
-bytemuck = {workspace = true, features = ["derive"]}
+bytemuck = { workspace = true, features = ["derive"] }
 
 [target.'cfg(any(target_arch = "aarch64", target_arch = "x86_64", target_arch = "x86"))'.dependencies]
 cpufeatures.workspace = true
@@ -16,10 +16,6 @@ rand_chacha.workspace = true
 rand.workspace = true
 rand_core.workspace = true
 criterion.workspace = true
-
-[features]
-armv8 = [] # Enable nightly-only ARMv8 intrinsics support
-force-soft = [] # Disable support for hardware intrinsics
 
 [[bench]]
 name = "clmul"
--- a/clmul/src/backend/autodetect.rs
+++ b/clmul/src/backend/autodetect.rs
@@ -97,31 +97,6 @@ impl Clmul {
         }
     }
 
-    /// Performs carryless multiplication. Same as clmul() but reusing the
-    /// operands to return the result. This gives a ~6x speed up compared
-    /// to clmul() where we create new objects containing the result.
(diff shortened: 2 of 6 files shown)


clmul/src/backend/autodetect.rs
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }
+    }
+}
+
+impl Clmul {
+    pub fn new(h: &[u8; 16]) -> Self {
+        let (token, has_intrinsics) = mul_intrinsics::init_get();
+
+        let inner = if cfg!(clmul_force_soft) || !has_intrinsics {
```
themighty1, line 56: if we don't have `intrinsics` and we didn't pass `clmul_force_soft`, this code will select `intrinsics` in the else{} branch below. But we need `soft` instead.

clmul/src/backend/autodetect.rs
```diff
@@ -0,0 +1,247 @@
...
+            )
+        }
+    }
+
+    /// Performs carryless multiplication. Same as clmul() but reusing the
+    /// operands to return the result. This gives a ~6x speed up compared
+    /// to clmul() where we create new objects containing the result.
+    /// The high bits will be placed in `self`, the low bits - in `x`.
+    #[inline]
+    pub fn clmul_reuse(&mut self, x: &mut Self) {
```
themighty1: Im in favor of getting rid of this fn since the speedup doesnt apply anymore.

clmul/Cargo.toml
```diff
@@ -17,10 +17,6 @@ rand.workspace = true
 rand_core.workspace = true
 criterion.workspace = true
 
-[features]
```
themighty1, line 20: There are some refs to `armv8` and `force-soft` dangling in the code and comments. 
Can we also document the new `clmul_armv8` and `clmul_force_soft` flags? We can document them here as a comment.

themighty1 (changes requested):
lgtm, a fix is needed

clmul/src/backend/autodetect.rs
sinui0: Are you sure? This evaluates true if
 - force soft is enabled, or
 - instrinsics aren't available

clmul/src/backend/autodetect.rs
sinui0: I agree

clmul/Cargo.toml
sinui0: I documented them in the crate root docs

clmul/src/backend/autodetect.rs
themighty1: you are right, sry

themighty1 (approved):
lgtm

PR #77, opened 2023-10-15
add SPCOT in mpz-ot-core

xiangxiecrypto:
1. Add the SPCOT protocol in the `mpz-ot-core` crate.
2. MPZ now lacks COT traits and implementations. Therefore, the current PR uses an ideal functionality of COT (with random choices).
3. The network IO is not involved.


sinui0:
Great, I'll review soon! Can you please rebase on `dev`?

xiangxiecrypto:
> Great, I'll review soon! Can you please rebase on `dev`?

Done

ot/mpz-ot-core/src/ferret/spcot/msgs.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+/// The message that receiver receives from the COT functionality.
+pub struct CotMsgForReceiver {
+    /// The random bits that receiver receives from the COT functionality.
+    pub rs: Vec<bool>,
+    /// The chosen blocks that receiver receivers from the COT functionality.
+    pub ts: Vec<Block>,
+}
```
sinui0: These messages would be part of the underlying `COT` protocol, we shouldn't define them here

ot/mpz-ot-core/src/ferret/spcot/msgs.rs
```diff
@@ -0,0 +1,66 @@
...
+#[allow(missing_docs)]
+pub enum Message<BaseMsg> {
+    BaseMsg(BaseMsg),
+    MaskBits(MaskBits),
+    ExtendFromSender(ExtendFromSender),
+    CotMsgForSender(CotMsgForSender),
+    CotMsgForReceiver(CotMsgForReceiver),
+    CheckFromReceiver(CheckFromReceiver),
+    CheckFromSender(CheckFromSender),
+}
```
sinui0: This protocol wraps a COT protocol, so we just add a variant for it. We can get rid of `BaseMsg` as that would be for a Base OT protocol which SPCOT doesn't use directly.

```suggestion
pub enum Message<CotMsg> {
    CotMsg(CotMsg),
    MaskBits(MaskBits),
    ExtendFromSender(ExtendFromSender),
    CheckFromReceiver(CheckFromReceiver),
    CheckFromSender(CheckFromSender),
}
```

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// # Arguments
+    ///
+    /// * `h` - The depth of the GGM tree.
+    /// * `alpha` - The chosen position.
+    /// * `extend` - The message from COT ideal functionality for the receiver. Only the random bits are used.
+    pub fn extend_mask_bits(
+        &mut self,
+        h: usize,
+        alpha: usize,
+        extend: CotMsgForReceiver,
```
sinui0:
```suggestion
        rs: &[bool],
```

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,293 @@
...
+    ///
+    /// * `h` - The depth of the GGM tree.
+    /// * `alpha` - The chosen position.
+    /// * `extendfc` - The message from COT ideal functionality for the receiver. Only the chosen blocks are used.
+    /// * `extendfr` - The message sent from the sender.
+    pub fn extend(
+        &mut self,
+        h: usize,
+        alpha: usize,
+        extendfc: CotMsgForReceiver,
```
sinui0: Instead of depending on a concrete message type here, we can instead just depend on the correlations directly.

```suggestion
        ts: &[Block]
```

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        // Computes V
+        let mut chis = vec![Block::ZERO; 1 << h];
+        Prg::from_seed(chis_seed).random_blocks(&mut chis);
+        v ^= Block::inn_prdt_red(&chis, &self.state.vs);
+
+        // Computes H'(V)
+        let mut hasher = blake3::Hasher::new();
+        hasher.update(&v.to_bytes());
+        let hashed_v = Hash::from(*hasher.finalize().as_bytes());
```
sinui0: We have a little helper function in `utils` which simplifies this, and `Hash` implements `From<[u8; 32]>`

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,227 @@
...
+        h: usize,
+        extend: CotMsgForSender,
+        mask: MaskBits,
+    ) -> Result<ExtendFromSender, SenderError> {
+        let CotMsgForSender { qs } = extend;
+        let MaskBits { bs } = mask;
+
+        if qs.len() != h {
+            return Err(SenderError::InvalidLength(
+                "the length of q should be 128".to_string(),
```
sinui0: `h` is an input argument and may not be 128. Is there a reason to not use a constant here?

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        // Step 3-4, Figure 6.
+
+        // Generates a GGM tree with depth h and seed s.
+        let s = self.state.prg.random_block();
+        let ggm_tree = GgmTree::new(h);
+        let mut k0 = vec![Block::ZERO; h];
+        let mut k1 = vec![Block::ZERO; h];
+        self.state.vs = vec![Block::ZERO; 1 << h];
+        ggm_tree.gen(s, &mut self.state.vs, &mut k0, &mut k1);
```
sinui0: Multiple calls to `extend` will overwrite `self.state.vs`

sinui0 (changes requested):
Nice! Looking good, but I have some suggestions around type/state safety.

Right now both the `Sender` and `Receiver` only have two states: `Initialized` and `Extend`. I think it would be best to lean more into the type state pattern to enforce correctness here. We should be able to statically guarantee the protocol executes in the correct sequence.

Separately, rather than coupling the sender and receiver to the `CotMsgFor*` messages, we should just pass in the correlations/choices directly as input arguments.

ot/mpz-ot-core/src/ferret/spcot/sender.rs
xiangxiecrypto: It is a typo, the length of `q` should be `h`

ot/mpz-ot-core/src/ferret/spcot/sender.rs
xiangxiecrypto: We should output `vs` after each call. 
Is it better that we make `vs` as the output of extend?

sinui0: `vs` should not be accessible until after the linear combination check has passed. This follows my suggestion at the top-level comment to introduce more states to make it impossible to misuse the API

sinui0: If you expect `extend` to be called multiple times prior to the check procedure, we can store them all internally until afterwards

xiangxiecrypto: Well, let me fix it, and probably optimize the performance.

xiangxiecrypto:
1. Allow multiple calls of the `extend` function before `check`.
2. Optimize with batch check.
3. Optimize with Fiat-Shamir.

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// See step 7 in Figure 6.
+    ///
+    /// # Arguments
+    ///
+    /// * `x_star` - The message from COT ideal functionality for the receiver. Only the random bits are used.
+    pub fn check_pre(&mut self, x_star: &[bool]) -> Result<CheckFromReceiver, ReceiverError> {
+        if x_star.len() != CSP {
+            return Err(ReceiverError::InvalidLength(
+                "the length of x* should be 128".to_string(),
```
sinui0:
```suggestion
                format!("the length of x* should be {CSP}")
```

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,316 @@
...
+        for (alpha, n) in &self.state.alphas_and_length {
+            let mut chis = vec![Block::ZERO; *n];
+            prg.random_blocks(&mut chis);
+            sum_chi_alpha ^= chis[*alpha];
+            self.state.chis.extend_from_slice(&chis);
+        }
+
+        let x_prime: Vec<bool> = sum_chi_alpha
+            .to_lsb0_vec()
+            .into_iter()
```
sinui0: we can avoid double allocation

```suggestion
            .iter_lsb0()
```

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        // Computes H'(W)
+        let hashed_w = Hash::from(blake3(&w.to_bytes()));
+
+        if hashed_v != hashed_w {
+            return Err(ReceiverError::ConsistencyCheckFailed);
+        }
+
+        self.state.cot_counter += self.state.unchecked_ws.len();
+        self.state.extended = true;
```
sinui0, line 251: Is this method intended to be terminating? If so, we can receive `self` instead of `&mut self` in the method signature

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,316 @@
...
+        impl Sealed for super::Initialized {}
+        impl Sealed for super::Extension {}
+    }
+
+    /// The receiver's state.
+    pub trait State: sealed::Sealed {}
+
+    /// The receiver's initial state.
+    #[derive(Default)]
+    #[allow(missing_docs)]
```
sinui0: unnecessary

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,262 @@
...
+            let tweaks = [tweak, tweak];
+            FIXED_KEY_AES.tccr_many(&tweaks, blks);
+        });
+
+        ms.iter_mut()
+            .zip(k0.iter().zip(k1.iter()))
+            .for_each(|([m0, m1], (k0, k1))| {
+                *m0 ^= *k0;
+                *m1 ^= *k1;
+            });
```
sinui0: Looks like we're iterating over `ms` twice, perhaps zip `k0` and `k1` into the first iterator?

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,262 @@
...
+        let hashed_v = Hash::from(blake3(&v.to_bytes()));
+
+        let mut res = Vec::new();
+        for n in &self.state.vs_length {
+            let tmp: Vec<Block> = self.state.unchecked_vs.drain(..*n).collect();
+            res.push(tmp);
+        }
+
+        self.state.cot_counter += self.state.unchecked_vs.len();
+        self.state.extended = true;
```
sinui0, line 195: Same here, we can terminate by receiving `self`

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -0,0 +1,316 @@
...
+                "the length of M should be h".to_string(),
+            ));
+        }
+
+        // Updates hasher
+        self.state.hasher.update(&ms.to_bytes());
+        self.state.hasher.update(&sum.to_bytes());
+
+        let mut alpha_bar_vec = alpha.to_msb0_vec();
+        alpha_bar_vec.drain(0..alpha_bar_vec.len() - h);
```
sinui0: `alpha` is `usize` which will be different depending on the architecture. Is this ok?

Also you can avoid the vector allocation with `alpha.msb0_iter().take(usize::BITS - h)`

ot/mpz-ot-core/src/ferret/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    #[test]
+    fn ideal_cot_test() {
+        let num = 100;
+        let mut ideal_cot = IdealCOT::init();
+        let (sender, receiver) = ideal_cot.extend(num);
+
+        assert!(ideal_cot.check(sender, receiver));
+    }
+}
```
sinui0: This is handy, can we move this to `src/ideal/cot` module? It can be used elsewhere. It would also be nice to have a `new_with_delta` method to set a predetermined delta

sinui0 (commented):
Nice work, a couple comments

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
xiangxiecrypto: Yes, but I think we still need to update the state, maybe for further usage. So I keep it as `&mut self`

ot/mpz-ot-core/src/ferret/spcot/receiver.rs
xiangxiecrypto: It should be ok to use `usize`, but I think fixing it as `u32` is safer.

should be `alpha.iter_msb0().skip(32 - h)`

ot/mpz-ot-core/src/ferret/mod.rs
xiangxiecrypto: good point

ot/mpz-ot-core/src/ferret/spcot/sender.rs
```diff
@@ -0,0 +1,262 @@
...
+                let tweak: Block = bytemuck::cast([i, self.state.exec_counter]);
+                let tweaks = [tweak, tweak];
+                FIXED_KEY_AES.tccr_many(&tweaks, blks);
+                blks
+            })
+            .zip(k0.iter().zip(k1.iter()))
+            .for_each(|([m0, m1], (k0, k1))| {
+                *m0 ^= *k0;
+                *m1 ^= *k1;
+            });
```
sinui0: I find this a bit hard to follow with the multiple closures, perhaps a simple loop?

```suggestion
        // Computes M0 and M1.
        let mut ms: Vec<[Block; 2]> = Vec::with_capacity(qs.len());
        for (((i, &q), b), (k0, k1)) in qs.iter().enumerate().zip(bs).zip(k0.into_iter().zip(k1)) {
            let mut m = if b {
                [q ^ self.state.delta, q]
            } else {
                [q, q ^ self.state.delta]
            };

            let tweak: Block = bytemuck::cast([i, self.state.exec_counter]);
            FIXED_KEY_AES.tccr_many(&[tweak, tweak], &mut m);

            m[0] ^= k0;
            m[1] ^= k1;

            ms.push(m);
        }
```

xiangxiecrypto: Well, I used iter twice before to make the code more readable.
For functionality, all these ways are workable.

sinui0: Sorry for being picky :grinning: The multiple closures iterates over `ms` 3 times I think (not sure if the compiler optimizes it away)

xiangxiecrypto: I think you are right on this point. Although I think the TCCR hash will dominate the performance, it is always good to make possible optimizations.

ot/mpz-ot-core/src/ideal/ideal_cot.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#[test]
+fn ideal_cot_test() {
+    let num = 100;
+    let mut ideal_cot = IdealCOT::init();
+    let (sender, receiver) = ideal_cot.extend(num);
+
+    assert!(ideal_cot.check(sender, receiver));
+}
```
sinui0: Please place into a `mod tests` module

sinui0 (approved):
Good work

themighty1 (commented):
lgtm, will open a new PR for some minor fixes.

PR #83, opened 2023-11-07
refactor: decouple zk execution from proving output

sinui0:
This PR decouples execution of a circuit from proving of the output in the ZK setting. This is needed to support incremental computation as we do with MPC. This also unblocks https://github.com/tlsnotary/tlsn/pull/379 which requires this feature.


th4s (approved):
:+1:

themighty1 (approved):
thanks, gw

I'll open a separate PR with some clarifying code comments once this PR is merged

PR #86, opened 2023-11-10
chore: misc comments for PR83

themighty1:
This PR adds comments to clarify:

- we are proving the authenticity of the circuit output
- other minor points

### DIFF
--- a/garble/mpz-garble/src/lib.rs
+++ b/garble/mpz-garble/src/lib.rs
@@ -370,22 +370,22 @@ pub trait Verify {
     ) -> Result<(), VerifyError>;
 }
 
-/// This trait provides methods for decoding output values.
+/// This trait provides methods for decoding values.
 #[async_trait]
 pub trait Decode {
-    /// Decodes the provided output values, returning the plaintext values to all parties.
+    /// Decodes the provided values, returning the plaintext values to all parties.
     async fn decode(&mut self, values: &[ValueRef]) -> Result<Vec<Value>, DecodeError>;
 }
 
-/// This trait provides methods for decoding output values with different privacy configurations.
+/// This trait provides methods for decoding values with different privacy configurations.
 #[async_trait]
 pub trait DecodePrivate {
-    /// Decodes the provided output values, returning the plaintext values to only this party.
+    /// Decodes the provided values, returning the plaintext values to only this party.
     async fn decode_private(&mut self, values: &[ValueRef]) -> Result<Vec<Value>, DecodeError>;
 
-    /// Decodes the provided output values, returning the plaintext values to the other party(s).
+    /// Decodes the provided values, returning the plaintext values to the other party(s).
     async fn decode_blind(&mut self, values: &[ValueRef]) -> Result<(), DecodeError>;
 
-    /// Decodes the provided output values, returning additive shares of plaintext values to all parties.
+    /// Decodes the provided values, returning additive shares of plaintext values to all parties.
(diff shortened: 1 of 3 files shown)


garble/mpz-garble/src/lib.rs
```diff
@@ -335,7 +335,8 @@ pub trait Execute {
     ) -> Result<(), ExecutionError>;
 }
 
-/// This trait provides methods for proving the output of a circuit.
+/// This trait provides methods for the evaluator to prove the authenticity of the evaluated garbled
+/// circuit's output.
```
sinui0, line 339: These traits are meant to be protocol generic, ie no mention of garbled circuits here. (I know its in the garble crate, but these will be moving)

garble/mpz-garble/src/lib.rs
```diff
@@ -361,30 +362,30 @@ pub trait Verify {
...
+    /// Verifies the provided output values against the expected values.
     async fn verify(
         &mut self,
         values: &[ValueRef],
         expected_values: &[Value],
     ) -> Result<(), VerifyError>;
 }
 
-/// This trait provides methods for decoding values.
+/// This trait provides methods for decoding output values.
```
sinui0: These don't have to be output values, input values can be decoded as well.

garble/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -332,7 +327,7 @@ impl DEAP {
         Ok(())
     }
 
-    /// Sends a commitment to the provided values, proving them to the follower upon finalization.
+    /// Sends a commitment to the provided output values, deferring the actual proving until finalization.
```
sinui0, line 330:
```suggestion
    /// Sends a commitment to the provided values, deferring the actual proving until finalization.
```

garble/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -354,7 +349,7 @@ impl DEAP {
         Ok(())
     }
 
-    /// Receives a commitment to the provided values, and stores it until finalization.
+    /// Receives a commitment to the provided output values, and stores it until finalization.
```
sinui0, line 352:
```suggestion
    /// Receives a commitment to the provided values, and stores it until finalization.
```

sinui0 (changes requested):

garble/mpz-garble/src/lib.rs
themighty1: Ok. Can we keep the garbling terminology now and when we move, we will adapt the wording to be generic?

garble/mpz-garble/src/lib.rs
themighty1: I can't think of any concrete scenarios where there would be a need to decode input values. I think we should try to keep the comments less abstract where possible (it will cause less abstraction fatigue for the reader).  Later, when there is a real use-case for using these traits on input values, we can modify the comments accordingly.

sinui0: We do it in @th4s latest PR to prove substrings of plaintext

themighty1: ah, ok, thx, goot to know. then i'll revisit this PR in light of that PR later on.

garble/mpz-garble/src/lib.rs
sinui0: I don't agree we should do that, the traits are already generic and I don't think coupling the trait docs to DEAP semantics helps any more than just understanding how DEAP implements them

themighty1:
N/A due to recent VM abstractions.

PR #87, opened 2023-11-22
add Mpcot

xiangxiecrypto:
Add two versions of mpcot protocols, taking spcot as an ideal functionality.
1. the chosen positions are under arbitrary distribution.
2. the chosen positions are under regular distribution, i.e., evenly distributed.


ot/mpz-ot-core/src/ideal/ideal_spcot.rs
```diff
@@ -0,0 +1,116 @@
...
+#[allow(missing_docs)]
+pub struct IdealSpcot {
+    pub delta: Block,
+    pub counter: usize,
+    pub prg: Prg,
+}
+
+impl IdealSpcot {
+    /// Initiate the functionality.
+    pub fn init() -> Self {
```
sinui0: This is typically named `new`

```suggestion
    pub fn new() -> Self {
```

ot/mpz-ot-core/src/ideal/ideal_spcot.rs
```diff
@@ -0,0 +1,116 @@
...
+        let delta = prg.random_block();
+        IdealSpcot {
+            delta,
+            counter: 0,
+            prg,
+        }
+    }
+
+    /// Initiate with a given delta
+    pub fn init_with_delta(delta: Block) -> Self {
```
sinui0:
```suggestion
    pub fn new_with_delta(delta: Block) -> Self {
```

ot/mpz-ot-core/src/ideal/ideal_spcot.rs
```diff
@@ -0,0 +1,116 @@
...
+            w_tmp[*alpha as usize] ^= self.delta;
+
+            v.push(v_tmp);
+            w.push(w_tmp);
+            self.counter += n;
+        }
+        (SpcotMsgForSender { v }, SpcotMsgForReceiver { w })
+    }
+
+    /// Performs the checks.
```
sinui0: Can you elaborate further on what "the checks" are? Perhaps this function is more appropriately a standalone test function?

ot/mpz-ot-core/src/ideal/ideal_spcot.rs
```diff
@@ -0,0 +1,116 @@
...
+pub struct SpcotMsgForReceiver {
+    /// The random blocks that receiver receives from the SPCOT functionality.
+    pub w: Vec<Vec<Block>>,
+}
+
+#[allow(missing_docs)]
+pub struct IdealSpcot {
+    pub delta: Block,
+    pub counter: usize,
+    pub prg: Prg,
```
sinui0: We should be able to keep these private

ot/mpz-ot-core/src/ferret/utils.rs
sinui0, line 1: Let's call this module `cuckoo`

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.buckets = vec![Vec::default(); self.m];
+        for i in 0..n {
+            for (index, hash) in self.hashes.iter().enumerate() {
+                let pos = hash_to_index(hash, self.m, i);
+                self.buckets[pos].push(Item {
+                    value: i,
+                    hash_index: index,
+                });
+            }
+        }
```
sinui0: Step 3.c specifies the bucket to be sorted ascending after insertion

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /// Inserts the input vector [0..n-1] into buckets.
+    ///
+    /// # Argument
+    ///
+    /// * `n` - The length of the vector [0..n-1].
+    #[inline]
+    pub fn insert(&mut self, n: u32) {
```
sinui0: Rather than performing this step separately, perhaps this calculation could be moved into `new`. Right now it allocates `self.bucket` twice. The `insert` method name implies that its a persistent collection but calling `insert` resets the buckets. It seems `Bucket` could be immutable instead

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+                self.buckets[pos].push(Item {
+                    value: i,
+                    hash_index: index,
+                });
+            }
+        }
+    }
+}
+
+// Always sets m = 1.5 * t. t is the length of `alphas`.
```
sinui0:
```suggestion
// Always sets m = 1.5 * t. t is the length of `alphas`. See Section 7.1 Parameter Selection
```

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// * Argument
+    ///
+    /// * `alphas` - A u32 vector being inserted.
+    #[inline]
+    pub fn insert(&mut self, alphas: &[u32]) -> Result<(), CuckooHashError> {
+        // Always sets m = 1.5 * t. t is the length of `alphas`.
+        self.m = compute_table_length(alphas.len() as u32);
+
+        // Allocates table.
+        self.table = vec![None; self.m];
```
sinui0: Similar to my comment on `Bucket`, this is a bit confusing as `insert` implies the table is persistent. We can make this struct immutable

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use super::{CUCKOO_HASH_NUM, CUCKOO_TRIAL_NUM};
+
+/// Errors that can occur when inserting Cuckoo hash.
+#[derive(Debug, thiserror::Error)]
+#[allow(missing_docs)]
+pub enum CuckooHashError {
+    #[error("invalid Cuckoo hash state: expected {0}")]
+    CuckooHashLoop(String),
+}
```
sinui0: We can simplify this as I believe there is only going to be 1 error case:

```suggestion
/// Cuckoo hash eviction error
#[derive(Debug, thiserror::Error)]
#[error("item evicted from table")]
pub struct Eviction;
```

ot/mpz-ot-core/src/ferret/utils.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Implementation of Bucket. See step 3 in Figure 7
+pub struct Bucket<'a> {
+    // The hash functions.
+    hashes: &'a [AesEncryptor; CUCKOO_HASH_NUM],
+    // The number of buckets.
+    m: usize,
+    /// The buckets contain all the elements.
+    pub buckets: Vec<Vec<Item>>,
```
sinui0: I think we can avoid leaking this

ot/mpz-ot-core/src/ferret/utils.rs
sinui0: I suspect you were running into some troubles with "self-referential structs" because both `CuckooHash` and `Bucket` store a reference to the hashes. This forced you to leak some implementation details which complicates the other modules a bit. To simplify this we can instead take the hash functions as an input parameter to the methods and make these immutable. This should let you treat both more as persistent structures with query APIs.

Alternatively, if you're concerned about the consistency of the hash functions between calls, you could instead store them in `CuckooHash` and `Bucket` as `Arc<[AesEncryptor; CUCKOO_HASH_NUM]>`. The performance penalty is likely negligible.

sinui0 (changes requested):
Looks good, and simpler than I thought. I'll stop this review here with the provided feedback and circle back once we encapsulate `CuckooHash` and `Bucket` a bit better. I think that will really help with duplication/readability.

ot/mpz-ot-core/src/ideal/ideal_spcot.rs
xiangxiecrypto: Let me update the comment, it checks the outputs satisfy the relation or not. It will only be used in tests.

ot/mpz-ot-core/src/ferret/utils.rs
xiangxiecrypto: I can remove the lifetime.
What is the major disadvantage of using mutable `CuckooHash`?

ot/mpz-ot-core/src/ferret/utils.rs
xiangxiecrypto: The sorted step can be removed, I checked with the authors, and my test shows that it is correct.

ot/mpz-ot-core/src/ferret/utils.rs
xiangxiecrypto: Only allocates once right? There is no allocation in `new`.

ot/mpz-ot-core/src/ferret/utils.rs
xiangxiecrypto: Can you elaborate further?

sinui0: Having this be pub and the Receiver interacting with it directly is a leaky abstraction. All bucket functionality including querying could be handled internally rather than exposing this field

ot/mpz-ot-core/src/ferret/utils.rs
sinui0: You're right about only 1 allocation, for some reason I was thinking that `Vec::new` allocates, but it doesn't.

ot/mpz-ot-core/src/ferret/utils.rs
sinui0: Ah ok, thanks. Can we leave a comment there for future reviewers?

ot/mpz-ot-core/src/ferret/utils.rs
sinui0: > What is the major disadvantage of using mutable CuckooHash?

As far as I can tell we never actually mutate it after construction. Immutable structures are easier to reason about

ot/mpz-ot-core/src/ferret/cuckoo.rs
```diff
@@ -0,0 +1,195 @@
...
+                item = x;
+                item.hash_index = (item.hash_index + 1) % CUCKOO_HASH_NUM;
+            } else {
+                // If no value assigned to position `pos`, end the process.
+                return Ok(());
+            }
+        }
+        Err(CuckooHashError)
+    }
+}
```
sinui0, line 91: This wasn't quite what I had in mind in last review, I was just pointing out that `CuckooHash` is not mutated and we can do something like this:

```suggestion
pub struct CuckooHash(Vec<Option<Item>>);

impl CuckooHash {
    /// Creates a new Cuckoo hash table.
    ///
    /// # Arguments
    ///
    /// * `hashes` - The hash functions.
    /// * `alphas` - A u32 vector being inserted.
    #[inline]
    pub fn new(
        hashes: &[AesEncryptor; CUCKOO_HASH_NUM],
        alphas: &[u32],
    ) -> Result<Self, CuckooHashError> {
        let m = compute_table_length(alphas.len() as u32);

        // Allocates table.
        let mut table = vec![None; m];
        // Inserts each alpha.
        for &value in alphas {
            Self::hash(hashes, &mut table, value)?
        }

        Ok(Self(table))
    }
    
    /// Returns the length of the table
    pub fn len(&self) -> usize {
      self.0.len()
    }

    /// Returns an iterator over the table.
    pub fn iter(&self) -> impl Iterator<Item = &Option<Item>> {
        self.0.iter()
    }

    // Hash an element to a position with the current hash function.
    #[inline]
    fn hash(
        hashes: &[AesEncryptor; CUCKOO_HASH_NUM],
        table: &mut [Option<Item>],
        value: u32,
    ) -> Result<(), CuckooHashError> {
        // The item consists of the value and hash index, starting from 0.
        let mut item = Item {
            value,
            hash_index: 0,
        };

        for _ in 0..CUCKOO_TRIAL_NUM {
            // Computes the position of the value.
            let pos = hash_to_index(&hashes[item.hash_index], table.len(), item.value);

            // Inserts the value to position `pos`.
            let opt_item = table[pos].replace(item);

            // If position `pos` is not empty before the above insertion, iteratively inserts the obtained value.
            if let Some(x) = opt_item {
                item = x;
                item.hash_index = (item.hash_index + 1) % CUCKOO_HASH_NUM;
            } else {
                // If no value assigned to position `pos`, end the process.
                return Ok(());
            }
        }
        Err(CuckooHashError)
    }
}
```

ot/mpz-ot-core/src/ferret/cuckoo.rs
```diff
@@ -0,0 +1,195 @@
...
+                let pos = hash_to_index(hash, self.m, i);
+                buckets[pos].push(Item {
+                    value: i,
+                    hash_index: index,
+                });
+            }
+        }
+        buckets
+    }
+}
```
sinui0, line 128: With buckets, I was suggesting we complete encapsulation like below so that we don't leak state into the sender and receiver implementations. This way all state and logic is neatly together:

```suggestion
pub struct Buckets {
    /// The hash functions.
    hashes: [AesEncryptor; CUCKOO_HASH_NUM],
    /// The length of the vector [0..n-1].
    n: u32,
    /// The buckets.
    buckets: Vec<Vec<Item>>,
    bucket_lengths: Vec<usize>,
}

impl Buckets {
    /// Creates new buckets.
    ///
    /// # Panics
    ///
    /// - Panics if the length of any bucket is greater than `usize::MAX / 2 - 1`.
    ///
    /// # Arguments
    ///
    /// * `hashes` - The Cuckoo table hash functions.
    /// * `m` - The number of buckets.
    /// * `n` - The length of the vector [0..n-1].
    #[inline]
    pub fn new(hashes: [AesEncryptor; CUCKOO_HASH_NUM], m: usize, n: u32) -> Self {
        let (buckets, bucket_lengths) = Self::new_buckets(&hashes, m, n);

        Self {
            hashes,
            n,
            buckets,
            bucket_lengths,
        }
    }

    /// Recomputes buckets with different inputs.
    /// 
    /// # Arguments
    /// 
    /// * `m` - The number of buckets.
    /// * `n` - The length of the vector [0..n-1].
    pub fn reset(&mut self, m: usize, n: u32) {
        let (buckets, bucket_lengths) = Self::new_buckets(&self.hashes, m, n);

        self.buckets = buckets;
        self.bucket_lengths = bucket_lengths;
        self.n = n;
    }

    fn new_buckets(hashes: &[AesEncryptor; CUCKOO_HASH_NUM], m: usize, n: u32) -> (Vec<Vec<Item>>, Vec<usize>) {
        let mut buckets = vec![Vec::default(); m];
        // NOTE: the sorted step in Step 3.c can be removed.
        for x in 0..n {
            for (i, hash) in hashes.iter().enumerate() {
                let pos = hash_to_index(hash, m, x);
                buckets[pos].push(Item {
                    value: x,
                    hash_index: i,
                });
            }
        }

        let bucket_lengths = buckets
            .iter()
            .map(|bucket| {
                (bucket.len() + 1)
                    .checked_next_power_of_two()
                    .expect("bucket length should be less than usize::MAX / 2 - 1")
                    .ilog2()
            })
            .collect();

        (buckets, bucket_lengths)
    }

    /// Returns the length of each bucket.
    pub fn bucket_lengths(&self) -> &[usize] {
        &self.bucket_lengths
    }

    /// Returns `p` in Step 4 of Figure 7.
    pub fn compute_p(&self, table: &CuckooHash) -> Vec<u32> {
        todo!()
    }

    /// Returns `r` in Step 5 of Figure 7.
    pub fn compute_r(&self, rt: &[Vec<Block>]) -> Result<Vec<Block>, BucketError> {
        todo!()
    }

    /// Returns `s` in Step 5 of Figure 7.
    pub fn compute_s(&self, st: &[Vec<Block>]) -> Result<Vec<Block>, BucketError> {
        todo!()
    }
}
```

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,210 @@
...
+        // Geneates the buckets.
+        let buckets = bucket.insert(n);
+
+        // Generates queries for SPCOT.
+        // See Step 4 in Figure 7.
+        let mut p = vec![];
+        for (alpha, bin) in table.iter().zip(buckets.iter()) {
+            if let Some(x) = alpha {
+                let pos = find_pos(bin, x)?;
+                if let Some(power) = (bin.len() + 1).checked_next_power_of_two() {
```
sinui0: It's not clear to me why we round the bucket lengths to be a power of 2. Is this mentioned in the paper somewhere?

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,210 @@
...
+                    p.push((power.ilog2() as usize, pos as u32));
+                    self.state.buckets_length.push(power);
+                } else {
+                    return Err(ReceiverError::InvalidBucketSize(
+                        "The next power of 2 of the bucket size exceeds the MAX number".to_string(),
+                    ));
+                }
+            } else if let Some(power) = (bin.len() + 1).checked_next_power_of_two() {
+                p.push((power.ilog2() as usize, bin.len() as u32));
```
sinui0: Here you use `bin.len()`, shouldn't this be `bin.len() + 1` according to Step 4?

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    /// Performs MPCOT extension.
+    ///
+    /// See Step 5 in Figure 7.
+    ///
+    /// # Arguments
+    ///
+    /// * `rt` - The vector received from SPCOT protocol on multiple queries.
+    /// * `n` - The total nunber of indices.
+    pub fn extend(&mut self, rt: &[Vec<Block>], n: u32) -> Result<Vec<Block>, ReceiverError> {
```
sinui0: Can we store `n` in the state so we eliminate the possibility of the user providing different values between `extend_pre` and `extend`?

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// Outputs the length of each bucket plus 1.
+    ///
+    /// See Step 1 to Step 4 in Figure 7.
+    ///
+    /// # Arguments
+    ///
+    /// * `alphas` - The queried indices.
+    /// * `n` - The total number of indices.
+    pub fn extend_pre(
+        &mut self,
```
sinui0: Is there a reason why we can't add another type state for this step? `state::PreExtend`? This will eliminate some possible error states. (Same comment for the other impls)

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        };
+
+        for len in self.state.queries_length.iter() {
+            if let Some(power) = len.checked_next_power_of_two() {
+                self.state.queries_depth.push(power.ilog2() as usize);
+            } else {
+                return Err(ReceiverError::InvalidInput(
+                    "The next power of 2 of each length exceeds the MAX number".to_string(),
+                ));
```
sinui0: Same here, can you elaborate on why this rounding is taking place? This error message will not provide much information to a user encountering it.

sinui0 (changes requested):

ot/mpz-ot-core/src/ferret/cuckoo.rs
sinui0: I'm not sure if `reset` is the right approach, we could stick with the `Arc<[AesEncryptor; ..]>` and regenerate buckets using `new`

ot/mpz-ot-core/src/ferret/cuckoo.rs
xiangxiecrypto: Well, I wanted to make the `insert` function explicitly to be consist with the definition. 
For the functionality, you can make it happen in the `new` function, but probably it is not friendly for reviewers to understand when referring to the paper.

ot/mpz-ot-core/src/ferret/cuckoo.rs
xiangxiecrypto: Similar reasons as I commented in the `CuckooHash` struct.

The functions between lines 173 and 185 should not be here, they are specific to the MPCOT protocol. We actually should make `CuckooHash` and `Bucket` independent of the MPCOT protocol. They can be used for other protocols such as PSI.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
xiangxiecrypto: The input length of SPCOT should be a power of 2.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
xiangxiecrypto: `bin.len()` is the position of last element, starting from `0`

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
xiangxiecrypto: Yes, we can

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
xiangxiecrypto: It is part of Extend, so I make it a sub-step of the Extension state.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
sinui0: > It is part of Extend, so I make it a sub-step of the Extension state.

From the paper it is the same state, but in our implementation it allows for misuse, eg calling `pre_extend` twice in a row which defeats the purpose of using type states. I don't see the downside of adding another type state here.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
sinui0: Can we just pad the query vector? I find this affects readability. Or is it necessary to do it in this way?

ot/mpz-ot-core/src/ferret/cuckoo.rs
sinui0: I personally think we can deviate from mirroring the paper exactly if it is to follow better programming practices. Of course, there is a limit to that, but in this case I don't think it would be harder to follow

ot/mpz-ot-core/src/ferret/cuckoo.rs
sinui0: Point taken, you're right those Ferret specific methods don't belong there. I'm in support of moving these to `mpz-core` if its simple enough and provides better encapsulation. My main concern is to not leak `Vec<Vec<Item>>` into the other structs.

xiangxiecrypto: I think we have to expose `Vec<Option<Item>>` and `Vec<Vec<Item>>` if cuckoo hash and bucket are taken as independent building blocks because they are not just for MPCOT, could be used for other cases.

So we can not encapsulate all these inside.

xiangxiecrypto:
We need to determine if we keep cuckoo hash and buckets stuck to MPCOT or make them independent.
if we prefer the first case, then I can encapsulate `Vec<Option<Item>>` and `Vec<Vec<Item>>` inside. Otherwise, we need to expose them.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -25,23 +25,20 @@ impl Receiver {
         }
     }
 
-    /// Completes the setup phase of the protocol.
+    /// Complets the setup phase for PreExtend.
```
sinui0:
```suggestion
    /// Completes the setup phase for PreExtend.
```

ot/mpz-ot-core/src/ferret/mpcot/sender.rs
```diff
@@ -25,30 +25,27 @@ impl Sender {
         }
     }
 
-    /// Completes the setup phase of the protocol.
+    /// Complets the setup phase for PreExtend.
```
sinui0:
```suggestion
    /// Completes the setup phase for PreExtend.
```

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -61,68 +58,76 @@ impl Receiver<state::Extension> {
     ///
     /// * `alphas` - The queried indices.
     /// * `n` - The total number of indices.
-    pub fn extend_pre(
+    #[allow(clippy::type_complexity)]
+    pub fn pre_extend(
         &mut self,
```
sinui0: This should take ownership of `self` instead of mutating in place.

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -61,68 +58,76 @@ impl Receiver<state::Extension> {
...
     /// Performs MPCOT extension.
     ///
     /// See Step 5 in Figure 7.
     ///
     /// # Arguments
     ///
     /// * `rt` - The vector received from SPCOT protocol on multiple queries.
-    /// * `n` - The total nunber of indices.
-    pub fn extend(&mut self, rt: &[Vec<Block>], n: u32) -> Result<Vec<Block>, ReceiverError> {
+    pub fn extend(&mut self, rt: &[Vec<Block>]) -> Result<Vec<Block>, ReceiverError> {
```
sinui0: Here again, to follow the typestate pattern, it should take ownership of self and return `Result<(Receiver<state::PreExtend>, Vec<Block>),_>`.

sinui0 (changes requested):
Thanks for adding the additional state. We just need to address the outstanding comments then we can merge :rocket: 

Regarding the encapsulation, I'm ok with putting that off to another PR so we can get this thing over the finish line

ot/mpz-ot-core/src/ferret/mpcot/receiver.rs
xiangxiecrypto: good point.

sinui0 (approved):
:muscle:

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
```diff
@@ -0,0 +1,192 @@
...
+            ));
+        }
+
+        // The range of each interval.
+        let k = (n + t - 1) / t;
+
+        let queries_length = if n % t == 0 {
+            vec![k as usize; t as usize]
+        } else {
+            let mut tmp = vec![k as usize; (t - 1) as usize];
```
themighty1, line 55: Hi, @xiangxiecrypto , I only just now got around to reviewing this PR.
Could you please clarify the following:

The paper says all intervals must be of the same size.
What is the justification of having the last interval be of a different size?
What if the last interval happens to be of size 1?
If there is a justification, could you formulate it, so I could add it here as a comment.
Or maybe we can just stick to always requiring that arguments (n,t) are such that `n % t == 0`?

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
```diff
@@ -0,0 +1,192 @@
...
+            tmp.push((n % k) as usize);
+            if tmp.iter().sum::<usize>() != n as usize {
+                return Err(ReceiverError::InvalidInput(
```
themighty1, line 58: Also, @xiangxiecrypto , I cant think of what values of (n,t) would cause this error to trigger. I think it never will.

themighty1 (commented):
Sorry for the belated review. @xiangxiecrypto , could you pls clarify a few things.

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
themighty1: Consider if n == 91 and t == 10 then the last interval will be 1, since k == 10.
But if k instead was set to 9, then all intervals (except the last one) would be of length 9 and the last one would be of length 10, which I think is what we want here.

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
themighty1: ah, nvm, (101, 10) will trigger this error

ot/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
xiangxiecrypto: In the current choosing parameters, we actually have `n % t == 0`. But in order to select more other parameters, I make it a little bit more flexible here.

xiangxiecrypto: In the following implementations, I will add a more general way to estimate different LPN parameters. If  all the parameters we choose satisfy `n % t == 0`, I will remove the other part.

PR #88, opened 2023-11-27
Add operators &, |, ! for Tracer<Bit>

voltrevo:
Enables you to write eg:

```rs
let a = builder.add_input::<bool>();
let b = builder.add_input::<bool>();
let a_or_b = a | b; // Same as `!(!a & !b)`
```

I wanted these ops when I made a circuit for [Rock Paper Scissors Lizard Spock](https://github.com/voltrevo/mpz/blob/a9d48fe/rock-paper-sls/src/build_circuit.rs#L30).

### DIFF
--- a/mpz-circuits/src/ops/binary.rs
+++ b/mpz-circuits/src/ops/binary.rs
@@ -1,4 +1,4 @@
-use std::ops::BitXor;
+use std::ops::{BitXor, BitAnd, Not, BitOr};
 
 use crate::{
     components::{Feed, Node},
@@ -212,7 +212,7 @@ pub(crate) fn inv_nbit<const N: usize>(
     std::array::from_fn(|n| state.add_inv_gate(a[n]))
 }
 
-impl<'a> BitXor<Tracer<'a, Bit>> for Tracer<'a, Bit> {
+impl<'a> BitXor for Tracer<'a, Bit> {
     type Output = Tracer<'a, Bit>;
 
     fn bitxor(self, rhs: Tracer<'a, Bit>) -> Self::Output {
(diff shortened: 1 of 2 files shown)


mpz-circuits/src/tracer.rs
```diff
@@ -50,3 +50,10 @@ where
         BinaryRepr::Array(tracer.into_iter().map(|tracer| tracer.into()).collect())
     }
 }
+
+impl<'a> Tracer<'a, Bit> {
+    /// Returns the single node associated with the bit.
+    pub fn node(&self) -> Node<Feed> {
+        self.to_inner().nodes()[0]
+    }
+}
```
voltrevo, line 63: So that `.to_inner().nodes()[0]` could be replaced with `.node()`, otherwise repeated many times now.

mpz-circuits/src/ops/binary.rs
```diff
@@ -212,19 +212,47 @@ pub(crate) fn inv_nbit<const N: usize>(
...
+}
+
+impl<'a> Not for Tracer<'a, Bit> {
+    type Output = Tracer<'a, Bit>;
 
-        let value = Bit::new([out]);
+    fn not(self) -> Self::Output {
+        let out = self.state.borrow_mut().add_inv_gate(self.node());
 
-        drop(state);
```
voltrevo, line 225: The `RefMut` only lives to the end of the method and the state isn't used before then, so there's no need to explicitly drop it.

sinui0 (approved):
Nice, thank you!

sinui0:
@voltrevo looks like you just need to run fmt

voltrevo:
> @voltrevo looks like you just need to run fmt

Done :)

PR #91, opened 2024-01-15
Add RandomOT traits and implement them for KOS15

th4s:
This PR adds the traits `RandomOTSender` and `ChosenMessageRandomOTReceiver` and implements them for KOS15.

### DIFF
--- a/ot/mpz-ot/src/kos/mod.rs
+++ b/ot/mpz-ot/src/kos/mod.rs
@@ -6,6 +6,8 @@ mod sender;
 
 pub use error::{ReceiverError, ReceiverVerifyError, SenderError};
 use futures_util::{SinkExt, StreamExt};
+use rand_chacha::ChaCha20Rng;
+use rand_core::{RngCore, SeedableRng};
 pub use receiver::Receiver;
 pub use sender::Sender;
 
@@ -35,6 +37,19 @@ pub(crate) fn into_base_stream<'a, St: IoStream<msgs::Message<T>> + Send + Unpin
         Ok(msg) => msg.try_into_base_msg().map_err(From::from),
         Err(err) => Err(err),
     })
+}
+
+/// Stretches a 16-byte seed to arbitrary length using ChaCha20.
+fn random_bytes_from_seed<const N: usize>(seed: [u8; 16]) -> [u8; N] {
+    let mut stretched_seed = [0_u8; 32];
+    let double_seed = [seed, seed].concat();
+
+    stretched_seed.copy_from_slice(&double_seed);
+    let mut rng = ChaCha20Rng::from_seed(stretched_seed);
+
+    let mut output = [0_u8; N];
+    rng.fill_bytes(&mut output);
+    output
 }
 
 #[cfg(test)]
--- a/ot/mpz-ot/src/kos/mod.rs
+++ b/ot/mpz-ot/src/kos/mod.rs
@@ -6,8 +6,6 @@ mod sender;
 
 pub use error::{ReceiverError, ReceiverVerifyError, SenderError};
 use futures_util::{SinkExt, StreamExt};
-use rand_chacha::ChaCha20Rng;
-use rand_core::{RngCore, SeedableRng};
 pub use receiver::Receiver;
 pub use sender::Sender;
 
@@ -37,19 +35,6 @@ pub(crate) fn into_base_stream<'a, St: IoStream<msgs::Message<T>> + Send + Unpin
         Ok(msg) => msg.try_into_base_msg().map_err(From::from),
         Err(err) => Err(err),
     })
-}
-
-/// Stretches a 16-byte seed to arbitrary length using ChaCha20.
-fn random_bytes_from_seed<const N: usize>(seed: [u8; 16]) -> [u8; N] {
-    let mut stretched_seed = [0_u8; 32];
-    let double_seed = [seed, seed].concat();
-
-    stretched_seed.copy_from_slice(&double_seed);
-    let mut rng = ChaCha20Rng::from_seed(stretched_seed);
-
-    let mut output = [0_u8; N];
-    rng.fill_bytes(&mut output);
-    output
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


ot/mpz-ot/src/kos/mod.rs
```diff
@@ -37,6 +39,19 @@ pub(crate) fn into_base_stream<'a, St: IoStream<msgs::Message<T>> + Send + Unpin
...
+    let mut stretched_seed = [0_u8; 32];
+    let double_seed = [seed, seed].concat();
+
+    stretched_seed.copy_from_slice(&double_seed);
+    let mut rng = ChaCha20Rng::from_seed(stretched_seed);
+
+    let mut output = [0_u8; N];
+    rng.fill_bytes(&mut output);
+    output
+}
```
sinui0: We now have an AES PRG in `mpz-core` that implements `rand` traits (and takes a Block/16-byte seed). Let's use that

ot/mpz-ot/src/lib.rs
```diff
@@ -90,6 +106,26 @@ where
...
+    T: Send + Sync,
+    U: Send + Sync,
+{
+    /// Outputs the choice bits and the corresponding messages.
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of random messages to receive.
+    async fn receive(&mut self, count: usize) -> Result<(Vec<T>, Vec<U>), OTError>;
+}
```
sinui0: We can just call this `RandomOTReceiver` and get rid of the warning. This is an interface and should not bring implementation details into consideration. It's on the user to understand the implications of how they are composing specific concrete implementations of these traits.

ot/mpz-ot/src/lib.rs
```diff
@@ -68,6 +68,22 @@ where
     ) -> Result<(), OTError>;
 }
 
+/// A random OT sender.
+///
+/// The OT sender gets two messages which are guaranteed to be random by the protocol.
```
sinui0: This trait promises no such properties about the messages other than they aren't chosen here. What distribution they are sampled from is up to the concrete implementation

ot/mpz-ot/src/lib.rs
```diff
@@ -68,6 +68,22 @@ where
...
+pub trait RandomOTSender<T>
+where
+    T: Send + Sync,
+{
+    /// Outputs two random messages.
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of random messages to get.
+    async fn random_messages(&mut self, count: usize) -> Result<Vec<T>, OTError>;
```
sinui0:
```suggestion
    async fn send_random(&mut self, count: usize) -> Result<Vec<T>, OTError>;
```

ot/mpz-ot/src/lib.rs
```diff
@@ -90,6 +106,26 @@ where
...
+where
+    T: Send + Sync,
+    U: Send + Sync,
+{
+    /// Outputs the choice bits and the corresponding messages.
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of random messages to receive.
+    async fn receive(&mut self, count: usize) -> Result<(Vec<T>, Vec<U>), OTError>;
```
sinui0:
```suggestion
    async fn receive_random(&mut self, count: usize) -> Result<(Vec<T>, Vec<U>), OTError>;
```

sinui0: IO Channels?

ot/mpz-ot/src/lib.rs
sinui0: IO Channels?

sinui0 (changes requested):
Cool, seems pretty straight forward.

Can we get a test in there too?

sinui0 (approved):
gw :+1:

ot/mpz-ot/src/kos/receiver.rs
```diff
@@ -350,6 +381,45 @@ where
...
+            .take_choices_and_keys();
+
+        Ok((
+            choices,
+            random_outputs
+                .into_iter()
+                .map(|block| {
+                    let mut prg = Prg::from_seed(block);
+                    let mut out = [0_u8; N];
+                    prg.fill_bytes(&mut out);
```
themighty1, line 415: we should use PRG only if `out.len()` > `block.len()`, otherwise we can just use `block` as is without the PRG, since it is already random.

ot/mpz-ot/src/kos/sender.rs
```diff
@@ -375,6 +398,42 @@ where
...
+            .state
+            .try_as_extension_mut()
+            .map_err(SenderError::from)?;
+
+        let random_outputs = sender.keys(count).map_err(SenderError::from)?;
+
+        let prng = |block| {
+            let mut prg = Prg::from_seed(block);
+            let mut out = [0_u8; N];
+            prg.fill_bytes(&mut out);
```
themighty1, line 425: same here

ot/mpz-ot/src/lib.rs
```diff
@@ -68,6 +68,30 @@ where
     ) -> Result<(), OTError>;
 }
 
+/// A random OT sender.
+#[async_trait]
+pub trait RandomOTSender<T>: ProtocolMessage
+where
+    T: Send + Sync,
+{
+    /// Outputs two random messages.
```
themighty1:
```suggestion
    /// Outputs pairs of random messages.
```

ot/mpz-ot/src/lib.rs
```diff
@@ -68,6 +68,30 @@ where
...
+    ///
+    /// # Arguments
+    ///
+    /// * `sink` - The IO sink to the receiver.
+    /// * `stream` - The IO stream from the receiver.
+    /// * `count` - The number of random messages to get.
```
themighty1:
```suggestion
    /// * `count` - The number of pairs of random messages to output.
```

themighty1 (changes requested):
we need to discuss if the use of PRG can be avoided

ot/mpz-ot/src/kos/receiver.rs
th4s: There is already another implementation for `<bool, Block>` instead of `<bool, [u8; N]>`.

What specific case do you have in mind? When `out.len() < block.len()` do you want to directly use the block and strip the bits accordingly?

themighty1: ah, i overlooked that <bool, Block> exists. no worries.

themighty1 (approved):

PR #98, opened 2024-02-01
refactor: chunk KOS extend message

### LINKED ISSUE #97: chunk KOS extension message
The `Extend` message sent from the KOS receiver to the sender during extension can be very large. This clashes with DoS mitigation at the IO layer which limits the size of incoming data.

https://github.com/privacy-scaling-explorations/mpz/blob/b8f94795f7f33936a0cb97515408f0d4bb0789db/ot/mpz-ot/src/kos/receiver.rs#L108

Instead of sending it as a single message, the receiver should send it in chunks with a max size.

sinui0:
This PR refactors KOS impl to chunk the extension message sent from the Receiver to the Sender.

Closes #97 and https://github.com/tlsnotary/tlsn-utils/issues/10

### DIFF
--- a/ot/mpz-ot-core/src/kos/mod.rs
+++ b/ot/mpz-ot-core/src/kos/mod.rs
@@ -28,7 +28,16 @@ pub(crate) type RngSeed = <Rng as SeedableRng>::Seed;
 /// AES-128 CTR used for encryption.
 pub(crate) type Aes128Ctr = ctr::Ctr64LE<aes::Aes128>;
 
-/// Returns the size in bytes of the extension message for a given number of OTs.
+/// Pads the number of OTs to accomodate for the KOS extension check and
+/// the extension matrix transpose optimization.
+pub fn pad_ot_count(count: usize) -> usize {
+    // Round up the OTs to extend to the nearest multiple of 64 (matrix transpose optimization).
+    let count = (count + 63) & !63;
+    // Add OTs for the KOS extension check.
+    count + CSP + SSP
+}
+
+/// Returns the size in bytes of the extension matrix for a given number of OTs.
 pub fn extension_matrix_size(count: usize) -> usize {
     count * CSP / 8
 }
--- a/ot/mpz-ot-core/src/kos/mod.rs
+++ b/ot/mpz-ot-core/src/kos/mod.rs
@@ -28,6 +28,11 @@ pub(crate) type RngSeed = <Rng as SeedableRng>::Seed;
 /// AES-128 CTR used for encryption.
 pub(crate) type Aes128Ctr = ctr::Ctr64LE<aes::Aes128>;
 
+/// Returns the size in bytes of the extension message for a given number of OTs.
+pub fn extension_matrix_size(count: usize) -> usize {
+    count * CSP / 8
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/ot/mpz-ot-core/src/kos/mod.rs
+++ b/ot/mpz-ot-core/src/kos/mod.rs
@@ -30,11 +30,11 @@ pub(crate) type Aes128Ctr = ctr::Ctr64LE<aes::Aes128>;
 
 /// Pads the number of OTs to accomodate for the KOS extension check and
 /// the extension matrix transpose optimization.
-pub fn pad_ot_count(count: usize) -> usize {
-    // Round up the OTs to extend to the nearest multiple of 64 (matrix transpose optimization).
-    let count = (count + 63) & !63;
+pub fn pad_ot_count(mut count: usize) -> usize {
     // Add OTs for the KOS extension check.
-    count + CSP + SSP
+    count += CSP + SSP;
+    // Round up the OTs to extend to the nearest multiple of 64 (matrix transpose optimization).
+    (count + 63) & !63
 }
 
 /// Returns the size in bytes of the extension matrix for a given number of OTs.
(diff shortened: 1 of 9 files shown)


th4s (commented):
Wouldn't it be cleaner to handle this in a general way in our messaging layer? Then this change would apply to every sink and stream we are using, while at the same time not mixing the protocol code with IO stuff.

I am thinking of something like wrapper functions in our `IoSink` and `IoStream` traits, which break down the payload if needed.

sinui0:
That thought certainly crossed my mind, but it didn't seem to be quite as straight forward.

1. The IO layer already breaks down the message into chunks. The limit we are running into is mostly regarding _deserialization_ and how much memory a received message is allowed to utilize. This limit needs to be specific to the application context and message type. We could add additional bounds to the IO traits to specify this, but that would make them harder to implement and use. This would be less of a problem with specialization, but that's not in the foreseeable future.
2. A less flexible middleground is to fork `tokio_util::codec::length_delimited::LengthDelimitedCodec` and make it aware of variant-level size limits, so we could specify a limit specifically for `Extend` at the codec level rather than baking this into the IO traits. I don't want to block increasing the TLSNotary protocol limit on this work.
3. I think at the end of the day the chunking will get baked into the callsite control flow anyhow. The optimal state of this code would be to *generate* the matrix in chunks, streaming them, and processing the chunks concurrently. That would require "chunk" message types anyhow. Given our deprecation plans, this is a simpler short-term remedy even if its not actually doing stream processing.

th4s:
Makes sense :+1:, I haven't put that much thought into it when I wrote that.

th4s (approved):
:+1:

ot/mpz-ot-core/src/kos/msgs.rs
```diff
@@ -33,15 +34,55 @@ impl<BaseMsg> From<MessageError<BaseMsg>> for std::io::Error {
...
+
+/// Extension message sent by the receiver.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Extend {
     /// The receiver's extension vectors.
     pub us: Vec<u8>,
 }
 
+impl Extend {
+    /// Splits the message into chunks.
```
themighty1:
```suggestion
    /// Returns an iterator over the chunks of the message.
```

ot/mpz-ot-core/src/kos/mod.rs
```diff
@@ -28,6 +28,11 @@ pub(crate) type RngSeed = <Rng as SeedableRng>::Seed;
 /// AES-128 CTR used for encryption.
 pub(crate) type Aes128Ctr = ctr::Ctr64LE<aes::Aes128>;
 
+/// Returns the size in bytes of the extension message for a given number of OTs.
```
themighty1:
```suggestion
/// Returns the size in bytes of the extension matrix for a given number of OTs.
```

ot/mpz-ot/src/kos/sender.rs
```diff
@@ -138,13 +140,36 @@ where
...
+
+        // Receive extension matrix from the receiver.
+        while extend.us.len() < expected_us {
+            let Extend { us: chunk } = stream
+                .expect_next()
+                .await?
+                .try_into_extend()
+                .map_err(SenderError::from)?;
+
+            extend.us.extend(chunk);
```
themighty1, line 172: Is this correct?
The receiver was splitting off from the end of the matrix, so the first chunk which arrives must go to the end of the reconstructed matrix, not to the beginning as it is here.

themighty1 (changes requested):

ot/mpz-ot/src/kos/sender.rs
sinui0: Woh, what... how do the unit tests pass? I'll look into it

sinui0: Good catch, the unit tests don't setup enough OTs to hit the chunk threshold.

sinui0: Fixed.

ot/mpz-ot/src/kos/receiver.rs
```diff
@@ -90,9 +93,11 @@ where
         let mut ext_receiver =
             std::mem::replace(&mut self.state, State::Error).try_into_extension()?;
 
+        let count = pad_ot_count(count);
+
         // Extend the OTs, adding padding for the consistency check.
         let (mut ext_receiver, extend) = Backend::spawn(move || {
-            let extend = ext_receiver.extend(count + CSP + SSP);
+            let extend = ext_receiver.extend(count);
```
themighty1, line 100: padding should not happen inside .extend() again

ot/mpz-ot/src/kos/receiver.rs
```diff
@@ -90,9 +93,11 @@ where
         let mut ext_receiver =
             std::mem::replace(&mut self.state, State::Error).try_into_extension()?;
 
+        let count = pad_ot_count(count);
+
         // Extend the OTs, adding padding for the consistency check.
```
themighty1: this comment is not accurate anymore

themighty1 (changes requested):

ot/mpz-ot/src/kos/receiver.rs
sinui0: I've removed the rounding that occurs in the core impl and replaced it with an input restriction to be a multiple of 64. This subtlety is not exposed to the user of the async wrapper.

themighty1 (approved):
gw

PR #103, opened 2024-02-20
Add (R)OLE to mpz

th4s:
This PR adds OLE and ROLE to mpz by introducing two new crates and creates the building blocks, like e.g. traits and module structure, for future OLE flavors and additional implementations.

~~I still need to do some small improvements here and there, but thought it is a good idea to get it out now.~~
Ready for review.


th4s:
This will need to be adapted/rebased to the new architecture for IO.

th4s:
Rebased and adapted to new IO design. Should be ready for review.

th4s:
Rebased onto `threading-refactor`. Ready for review.

crates/mpz-ole-core/src/ideal/ole.rs
```diff
@@ -0,0 +1,110 @@
+//! This module provides an ideal OLE functionality.
+
+use mpz_fields::Field;
+use rand::{rngs::ThreadRng, thread_rng};
+
+/// The OLE functionality
+pub struct OLEFunctionality<F> {
```
sinui0: I recommend sticking to `Ideal*`, "functionality" is a mouthful for a type name.

crates/mpz-ole-core/src/ole/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Provides implementations of OLE with errors (OLEe) protocols.
```
sinui0:
```suggestion
//! Oblivious linear evaluation protocols
```

The "with errors" should be a caveat documented on the specific implementation, not at the module level

crates/mpz-ole-core/src/ideal/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! This module provides ideal functionalities for OLE and ROLE.
```
sinui0:
```suggestion
//! Ideal OLE functionalities.
```

They're all just variants of OLE, we can keep the module doc general

crates/mpz-ole-core/src/ideal/role.rs
```diff
@@ -0,0 +1,94 @@
+//! This module provides an ideal ROLE functionality.
```
sinui0:
```suggestion
//! Ideal ROLE.
```

Style wise, I think I prefer very succinct module and type headlines. It just reads cleaner with rustdoc to avoid leading words like "this module provides..." etc. Of course, its totally fine to be more verbose in the following description

crates/mpz-ole-core/src/ideal/role.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use mpz_fields::Field;
+use rand::{rngs::ThreadRng, thread_rng};
+
+/// The ROLE functionality
```
sinui0:
```suggestion
/// Ideal ROLE.
```

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+#![deny(missing_docs, unreachable_pub, unused_must_use)]
+#![deny(unsafe_code)]
+#![deny(clippy::all)]
+
+//! This crate provides implementations of different Oblivious Linear Evaluation with Errors (OLEe)
+//! flavors. It provides the core logic of the protocols without I/O.
```
sinui0:
```suggestion
//! Core implementations of various Oblivious Linear Evaluation protocols.
```

themighty1:
Im happy to review this once we all agree on what specific approach we will be taking for E2F and GHASH. For now, we only tentatively agreed to use the PADO's approach with modifications to work around the 0 input issue.
I have not examined PADO's protocols in detail yet.
So, Im expecting the following sequence of events:
- Jordan ACKs that he's ok with using PADO's E2F and GHASH with our 0-input fix
- I'll review PADO's E2F and GHASH to make sure they make sense
- I'll analyze the ROT->ROLE protocol again with greater scrutiny
- I'll review the rest of this PR

sinui0:
I haven't gone very deep in review here yet, but lets address some high level points:

1. I don't think it is appropriate to describe this protocol as Random OLE. That would imply the protocol deals random shares to both parties, which is not the case here.
2. Looking at your write up, it appears that "Random OLE" is exactly MASCOT (COPEe KOS16) with an additional step where the parties adjust their shares. AFAICT this share adjustment is superfluous, as they already choose them once and can sample them randomly if that is what they want, eg for offline preprocessing. I suggest we remove this step and just implement KOS16 verbatim.

    It recently occurred to me that the best approach is likely to just re-use the PRG seeds shared during the base OT for OT extension. Using the messages from ROT to seed the PRGs in COPEe seems unnecessary, and prevents us from preprocessing the OLEs in parallel with the OT extension.

3. We should stick with common nomenclature: OLE Sender and Receiver. Renaming these parties to Provider and Evaluator is bound to cause confusion. To help grok these terms: The Sender sends a linear function to the protocol and the Receiver receives an evaluation of that function at the index of their choice.
4. Regarding the "0-input fix" that should exist outside this OLE implementation entirely. It would be a sub-protocol of GHASH. Let's discuss that later, elsewhere.
5. The "with errors" suffix should not be present in any of the traits. We should just have an `OleSender` and `OleReceiver` trait, and users must understand the real functionality of the protocol they use which implements these traits.

crates/mpz-ole-core/src/role/ot/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// This is used to check at compile-time that the correct const-generic implementation is used for
+/// a specific field.
+struct Check<const N: usize, F: Field>(std::marker::PhantomData<F>);
+
+impl<const N: usize, F: Field> Check<N, F> {
+    const IS_BITSIZE_CORRECT: () = assert!(
+        N as u32 == F::BIT_SIZE / 8,
+        "Wrong bit size used for field. You need to use `F::BIT_SIZE` for N."
+    );
+}
```
sinui0: We should just add an associated const `BYTE_SIZE` to `Field`

crates/mpz-ole-core/src/role/mod.rs
sinui0, line 1: Let's remove this entire module and focus on OLEe

sinui0 (changes requested):
I had some WIP review here that I'll just post in addition to my previous comment.

Additionally, I was expecting to see the use of a PRF in the OLEe implementation, but saw it no where? It looked like you are operating directly on the ROT messages instead of using them as keys to the PRF.

I think we can avoid any mention of ROT in this crate, we simply use ROT to sample the PRF keys for KOS16.

crates/mpz-ole-core/src/ole/mod.rs
themighty1: While we're at it, we can be more verbose in our comments "oblivious linear evaluation" is also a short-hand which may be confusing to the uninitiated. The full term "oblivious linear **function** evalution", i.e. there exists a linear function which gets evaluated obliviously.

th4s:
Thanks for the review.

1) Can you explain why you think that the parties are not dealt random shares?

2) The difference between KOS16 (which is a VOLE) and this ROLE implementation is that in KOS16 the Delta is fixed. So the "Extend" calls cannot be used as a basis for an OLE, because only one party can choose the input. But for E2F or Ghash we need this degree of freedom. Am I missing something?

3) Agree

4) The 0-fix is not implemented. It only exists in the research write-up and we probably won't need it.

5) Agree

th4s:
I went the route of using PRNGs for ROLE. So that you also have a Setup call and Extend calls. I don't think that this works. Relevant PR https://github.com/tlsnotary/docs-mdbook/pull/65

crates/mpz-ole-core/src/role/mod.rs
th4s: The nice thing of going from ROLE -> OLE is that ROLE can be preprocessed. If we do not use ROLE to instantiate OLE then we probably go with Gilboa99, which does not allow preprocessing.

sinui0:
> Can you explain why you think that the parties are not dealt random shares?

In steps 1 and 2 of your `Extend` the parties sample their target shares, then subsequently use them to adjust the VOLE from COPEe. The functionality is not dealing these shares, the parties are choosing them. There is no constraint on the distribution from which these shares are sampled. I suggest we just skip this step, which is equivalent to COPEe.

> The difference between KOS16 (which is a VOLE) and this ROLE implementation is that in KOS16 the Delta is fixed. So the "Extend" calls cannot be used as a basis for an OLE, because only one party can choose the input. But for E2F or Ghash we need this degree of freedom. Am I missing something?

KOS16 indeed does implement VOLE, or in our case $l = 1$ which is a single OLE. You're correct though, we can't reuse the base OTs as I was suggesting earlier, so we can stick with using ROT from OT extension to sample/transfer the PRG seeds.

I'm not quite sure that your protocol is secure. Re-iterating, it looks to me to be COPEe where each extend call to COPEe is then followed by a share adjustment to convert the VOLE into an OLE. This would imply you can get n-OLE from VOLE simply with this share adjustment. Maybe that's true, but it's not self-evident to me that it's secure.

th4s:
> In steps 1 and 2 of your Extend the parties sample their target shares, then subsequently use them to adjust the VOLE from COPEe. The functionality is not dealing these shares, the parties are choosing them. There is no constraint on the distribution from which these shares are sampled. I suggest we just skip this step, which is equivalent to COPEe.

> I'm not quite sure that your protocol is secure. Re-iterating, it looks to me to be COPEe where each extend call to COPEe is then followed by a share adjustment to convert the VOLE into an OLE. This would imply you can get n-OLE from VOLE simply with this share adjustment. Maybe that's true, but it's not self-evident to me that it's secure.

I am not sure what you mean by `Extend`. For ROLE in the current construction (https://github.com/tlsnotary/docs-mdbook/blob/main/research/ole-flavors.typ#L29-L43) there is no `Extend` call. The construction  is pretty different from COPEe, which was only the starting point.


> KOS16 indeed does implement VOLE, or in our case which is a single OLE. You're correct though, we can't reuse the base OTs as I was suggesting earlier, so we can stick with using ROT from OT extension to sample/transfer the PRG seeds.

When I designed this, I tried to come up with with a `Setup, Extend` approach for ROLE, too, but couldn't get this to work in a way which made me feel safe about it. I did not invest too much time into this, because we have preprocessed random OTs available, so I felt that the performance impact was negligible. I tried using PRG seeds for ROLE, but I was not sure that this is safe.

crates/mpz-ole-core/src/cope/receiver.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// * `t_delta_i` - The receiver's random OT messages.
+    /// * `ui` - The correlations, sent by the sender.
+    ///
+    /// # Returns
+    ///
+    /// * `qk` - The receiver's final OLE output summands.
+    pub fn generate(
+        &self,
+        delta_k: &[F],
+        t_delta_i: &[[u8; N]],
```
sinui0: `t_delta_i` can instead be `[F]` and its up to the user to provide it in that format. We can add the bit iterator trait as a bound on F (not on the Field trait itself).

crates/mpz-ole-core/src/cope/receiver.rs
```diff
@@ -0,0 +1,71 @@
...
+        if delta_k.len() * F::BIT_SIZE as usize != t_delta_i.len() || t_delta_i.len() != ui.len() {
+            return Err(OLECoreError::LengthMismatch(format!(
+                "Number of choices {}, received OT messages {} and received correlations {} are not equal.",
+                delta_k.len(),
+                t_delta_i.len(),
+                ui.len(),
+            )));
+        }
+
+        let delta_i: Vec<bool> = delta_k.iter_lsb0().collect();
```
sinui0: I believe we can avoid this extra allocation using iterator combinators

crates/mpz-ole-core/src/cope/sender.rs
```diff
@@ -0,0 +1,73 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `ti01` - The OT messages, which the sender has sent to the receiver.
+    /// * `xk` - The sender's inputs to the OLE.
+    ///
+    /// # Returns
+    ///
+    /// * `ui` - The correlations, which will be sent to the receiver.
+    /// * `t0k` - The sender's final OLE output summands.
```
sinui0: These are simply "the Sender's shares", maybe `ak`,  right?

crates/mpz-ole-core/src/ideal.rs
sinui0, line 1: Rather than having multiple methods here which mutable some internal state of the functionality, I think it can simply provide a method like:

```rust
pub fn execute<F: Field>(
  &mut self, 
  sender_input: Vec<F>, 
  receiver_input: Vec<F>
) -> (Vec<F>, Vec<F>)
```

sinui0 (commented):
Algorithms LGTM :+1: 

I would suggest that you refactor a bit more to remove the `*Sender` and `*Receiver` types, as they have no internal state and complicate the API. Instead of an actor operating on data, we should instead provide newtypes for the data and implement the API directly on them. For example:

```rust
pub struct SenderShare<F>(F);

impl<F: Field> SenderShare<F> {
  pub fn new(share: F, corr: [[F; 2]; F::BIT_LENGTH]) -> (Self, MaskedInput<F>) {
    ...
  }
  
  pub fn adjust(self, target: F) -> (SenderAdjust<F>, ShareAdjust<F>) {
    ...
  }
}

pub struct ReceiverShare<F>(F);

impl<F: Field> ReceiverShare<F> {
  pub fn new(
    share: F,
    sender_input: MaskedInput<F>,
    corr: [F; F::BIT_LENGTH]
  ) -> Self {
    ...
  }
  
  pub fn adjust(self, target: F) -> (ReceiverAdjust<F>, ShareAdjust<F>) {
    ...
  }
}

pub struct ShareAdjust<F>(F);

pub struct SenderAdjust<F>(F);

impl<F: Field> SenderAdjust<F> {
  pub fn finish(self, receiver_adjust: ShareAdjust<F>) -> SenderShare<F> {
    ...
  }
}

pub struct ReceiverAdjust<F>(F);

impl<F: Field> ReceiverAdjust<F> {
  pub fn finish(self, sender_adjust: ShareAdjust<F>) -> ReceiverShare<F> {
    ...
  }
}
```

This would just be the core API, we can still add the sender and receiver types to store the preprocessed OLEs

crates/mpz-ole-core/src/cope/sender.rs
th4s: I tried to use the MASCOT paper's notation for this module, so that understanding it and checking for correctness is as easy as possible.

th4s:
Alright, I tried it out and it seems to work out quite nicely.

th4s:
Closing in favor of #135.

PR #107, opened 2024-03-05
feat: mpz-common

sinui0:


mpz-common/src/executor/st.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// # Arguments
+    ///
+    /// * `io` - The I/O channel used by the executor.
+    pub fn new(io: Io) -> Self {
+        Self {
+            id: ThreadId::default(),
+            io,
+        }
+    }
+}
```
th4s, line 29: This is somehow assuming that there will be only one `STExecutor`, because otherwise  they will have the same `ThreadId`.  Maybe we should either delegate responsibility to the caller and make `ThreadId` an argument of `new` or use some Singleton pattern?

mpz-common/src/executor/st.rs
```diff
@@ -0,0 +1,122 @@
...
+#[cfg(test)]
+mod tests {
+    use futures::executor::block_on;
+    use scoped_futures::ScopedFutureExt;
+    use serio::channel::duplex;
+
+    use super::*;
+
+    #[derive(Debug, Default)]
+    struct Test {
```
th4s: Maybe rename to `LifetimeTest`

mpz-common/src/id.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Returns the thread ID as a byte slice.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Increments the thread ID, returning `None` if the ID overflows.
+    #[inline]
+    pub fn increment(&self) -> Option<Self> {
```
th4s, line 30: When will this function be used? My understanding was that new threads are created via `fork`.

mpz-common/src/sync/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+        R: Unpin,
+    {
+        let tick = io.expect_next().await?;
+        Ok(Wait::new(&self.queue, tick, f).await)
+    }
+}
+
+#[derive(Debug, Default)]
+struct Queue {
+    tick: Ticket,
```
th4s, line 128: This being a ticket did not click for me. It is the current head of the queue, right? Can we somehow add this context information?

th4s (approved):
Very nice :rocket:

Took me some time to digest the sync mechanism ^^. So basically we synchronize the execution of green threads across machines using the ticketing mechanism.

If a green thread with the wrong ticket wants to run, it is put to sleep and added to the `Queue`. And the function they execute in most cases is handing out a lock to a resource.

Questions:
1) Will we have a multi-threaded runtime. Will this be a new PR?
2) What is the connection between `Mutex` and `Executor`? Will there be some parent struct having an `Executor` and `Mutex`es for several resources.

sinui0:
> Will we have a multi-threaded runtime. Will this be a new PR?

Yes, in a coming PR. We need it to get the GC VM working again

> What is the connection between Mutex and Executor? Will there be some parent struct having an Executor and Mutexes for several resources.

They aren't directly related. The `Mutex` is just a utility that can be slapped around a protocol when it needs to be shared. Ideally it won't be needed for very many things, mostly just OT and VOLE/PCG.

Side note: I'm not sure if `Executor` is the right name, it doesn't actually execute anything. Might end up renaming it when it becomes a bit more clear how the model pans out

mpz-common/src/executor/st.rs
sinui0: There is only 1 `STExecutor` per connection, I don't see the need to be able to specify the thread ID. Single threaded means 1 thread. If they want more than 1 then they should be using the coming multithreaded stuff

mpz-common/src/id.rs
sinui0: Fork adds a level to the chain, increment increases the id. Eg. fork(0) -> 0/0, fork(0/0) -> 0/0/0, increment(0/0/0) -> 0/0/1

sinui0: Later we'll add the ability to join multiple tasks, so we need to be able to fork N child threads

mpz-common/src/sync/mod.rs
sinui0: yes, its the current ticket which the queue will accept.

mpz-common/src/executor/st.rs
th4s: I had a misunderstanding of having only one single `Executor`. Makes sense.

mpz-common/src/sync/mutex.rs
```diff
@@ -0,0 +1,107 @@
...
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use super::*;
+
+    #[test]
+    fn test_mutex_st() {
+        let leader_mutex = Arc::new(Mutex::new_leader(()));
+        let follower_mutex = Arc::new(Mutex::new_follower(()));
```
themighty1, line 88: I was confused here. If mutex synchronizes exclusive access to a resource, shouldn't Mutex::new_leader and Mutex::new_follower be given the same resource (or rather the pointer to the same resource)? 
Here each of them is given a distinct instance of ()

themighty1 (approved):
nice work, one q

mpz-common/src/sync/mutex.rs
sinui0: For a concrete example, one party would have `Arc<Mutex<KosSender>>` and the other `Arc<Mutex<KosReceiver>>`. The mutexes exist on different machines, and also sometimes you want to synchronize access to different types.

PR #109, opened 2024-03-05
refactor: mpz-ot

sinui0:
This PR builds on #107 and #108 , refactoring our OT traits and implementations. It also _deletes_ the KOS actor now that we have the `mpz_common::sync::Mutex` primitive :pray: 

I haven't finished with the committed/verifiable functionalities on the shared KOS variants yet, but will punt that to another PR so we can just move on for now.


ot/mpz-ot/src/kos/shared_receiver.rs
```diff
@@ -0,0 +1,55 @@
...
+use utils_aio::non_blocking_backend::{Backend, NonBlockingBackend};
+
+use crate::{
+    kos::{Receiver, ReceiverError},
+    OTError, OTReceiver,
+};
+
+/// A shared KOS receiver.
+#[derive(Debug, Clone)]
+pub struct SharedReceiver<BaseOT> {
```
th4s, line 17: Shouldn't these follow the new `Syncer` and `Mutex` approach?

th4s (approved):
gw :+1:

ot/mpz-ot/src/kos/shared_receiver.rs
th4s: I realized now that they do :sweat_smile: 

Can we eventually rename the `Mutex`. I am open to suggestions. What about `MpcMutex`?

Btw. this is an awesome abstraction :)

sinui0: I like just calling it `Mutex`, prefixing it is a bit redundant. If it raises a reviewers eyebrow they just need to look that its imported from `mpz_common` and is documented :)

sinui0:
Proceeding with merge as this should not introduce security issues, but @themighty1 if you get a time to sanity check where I've touched the coin-toss protocols in CO15 and KOS15 that would be appreciated :pray:

PR #114, opened 2024-03-08
feat(mpz-garble): add getters for thread in pool

sinui0:
This PR adds getters for the first thread in a pool, and enforces that a pool can not be empty.


th4s (approved):
:+1:

PR #115, opened 2024-03-11
Adds shared implementations for ROLE

th4s:
This PR adds `SharedROLEeProvider` and `SharedROLEeEvaluator` which offer preprocessed ROLEs and can make use of `mpz_common::sync::Mutex`


th4s:
Will be replaced by new PR.

PR #119, opened 2024-04-20
Fix AND gate unit test 

### LINKED ISSUE #118: bug(mpz-garble-core): unit test does not test 0-0 case
https://github.com/privacy-scaling-explorations/mpz/blob/eed88d141b8cc81e58ddac20abffa70fc4481d29/garble/mpz-garble-core/src/lib.rs#L100

Our AND gate unit test does not test the 0-0 input case for a gate.

caglaryucekaya:
Fixes #118


sinui0 (approved):
Thanks!

PR #122, opened 2024-04-30
refactor(mpz-ot): Normalize OT and ideal functionalities

sinui0:
This PR normalizes our OT interfaces as well as completely reimplements our ideal functionality abstraction.

Nothing here touches the security critical logic of any of the OT implementations. This is purely a clean up and testing utilities.

# Changes

- Ideal functionalities logic is now consolidated into `mpz-ot-core` instead of leaking into `mpz-ot`.
- Added an abstraction which makes it easier to implement ideal 2-party functionalities in `mpz-common`
- Every oblivious transfer call now has a `TransferId`
- Outputs for each OT functionality is now strongly typed.


crates/mpz-common/src/ideal.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(Debug, Default)]
+struct Buffer {
+    alice: HashMap<ThreadId, (BoxAny, oneshot::Sender<BoxAny>)>,
+    bob: HashMap<ThreadId, (BoxAny, oneshot::Sender<BoxAny>)>,
+}
+
+/// The ideal functionality from the perspective of Alice.
+#[derive(Debug)]
+pub struct Alice<F> {
+    f: Arc<Mutex<F>>,
```
th4s, line 23: What exactly is `f` here? Is it potential state of the functionality? Having a look at `ideal_f2p` it seems like Alice and Bob share the same `f`. So I do not fully understand what you mean by `The ideal functionality from the perspective of ...`.

crates/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -140,10 +145,18 @@ impl Receiver<state::Setup> {
...
+            ..
         } = &mut self.state;
 
-        let SenderPayload { payload } = payload;
+        let SenderPayload { id, payload } = payload;
+
+        // Check that the transfer id matches
+        let expected_id = current_id.next();
+        if id != expected_id {
+            return Err(ReceiverError::IdMismatch(id, expected_id));
```
th4s: This is in the wrong order.
```suggestion
            return Err(ReceiverError::IdMismatch(expected_id, id));
```

crates/mpz-ot-core/src/chou_orlandi/sender.rs
```diff
@@ -124,11 +128,21 @@ impl Sender<state::Setup> {
...
-        let ReceiverPayload { blinded_choices } = receiver_payload;
+        let ReceiverPayload {
+            id,
+            blinded_choices,
+        } = receiver_payload;
+
+        // Check that the transfer id matches
+        let expected_id = current_id.next();
+        if id != expected_id {
+            return Err(SenderError::IdMismatch(id, expected_id));
```
th4s:
```suggestion
            return Err(SenderError::IdMismatch(expected_id, id));
```

crates/mpz-ot-core/src/ideal/mpcot.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn set_delta(&mut self, delta: Block) {
+        self.delta = delta;
+    }
+
+    /// Performs the extension of MPCOT.
+    ///
+    /// # Argument
+    ///
+    /// * `alphas` - The positions in each extension.
+    /// * `n` - The length of the vector.
```
th4s, line 44: `t` is missing here.

crates/mpz-ot-core/src/ideal/spcot.rs
th4s, line 1: Is there a good reason to have SPCOT functionality implemented in batch mode, while MPCOT functionality is only a single execution?

crates/mpz-ot/src/ideal/mod.rs
```diff
@@ -2,5 +2,5 @@
 
 pub mod cot;
 pub mod ot;
-pub mod rcot;
-pub mod rot;
+// pub mod rcot;
+// pub mod rot;
```
th4s: Is this a left-over or will this be a follow-up PR?

th4s (approved):
Nice work :rocket:

What is the benefit of adding an extra encapsulation layer for the OT return types in the traits?

crates/mpz-common/src/ideal.rs
sinui0: > Is it potential state of the functionality?

Yes, its just a struct with an API for executing the functionality. They both hold a reference to it. It is "from the perspective of" because they are providing different inputs to it.

You can think of `F` as being `Fn(AliceInput, BobInput) -> (AliceOutput, BobOutput)`. The `call` method on `Alice` provides `AliceInput`, the `call` method on `Bob` provides `BobInput`.

crates/mpz-ot-core/src/ideal/spcot.rs
sinui0: I was just retrofitting what Xiang implemented. We'll probably want to refactor it at some point.

crates/mpz-ot/src/ideal/mod.rs
sinui0: woops, no these should be implemented

sinui0:
> What is the benefit of adding an extra encapsulation layer for the OT return types in the traits?

It just provides stronger typing and a way to document what each field is. It's better to have the types defined in `mpz-ot-core` and the traits in`mpz-ot` work with them

crates/mpz-ot/src/ideal/rot.rs
```diff
@@ -1,238 +1,78 @@
-//! Ideal functionality for random oblivious transfer.
+//! Ideal functionality for correlated oblivious transfer.
```
th4s:
```suggestion
//! Ideal functionality for random oblivious transfer.
```

th4s (approved):

PR #123, opened 2024-05-02
feat: generic circuit struct

brech1:
# Description

This PR aims to introduce a new generic circuit struct:

```rust
pub struct Circuit<T> {
    gates: Vec<T>,
}
```

The `Circuit` is a collection of gates. Since for the execution of these gates is important that these form a directed acyclic graph (DAG), the correct way of building this circuit is using the `CircuitBuilder`

```rust
pub struct CircuitBuilder<T>
where
    T: Component,
{
    gates: Vec<T>,
    input_map: HashMap<usize, usize>,
}
```

The `CircuitBuilder` sort the gates topologically, this means in the order they should be executed taking dependencies into account, and ensures that they form a DAG.

Here is our only constrain for the gates, they have to implement the `Component` trait, this means that we should be able to fetch their inputs and outputs. This is strictly necessary to sort the gates.

```rust
pub trait Component {
    fn get_inputs(&self) -> Vec<usize>;
    fn get_outputs(&self) -> Vec<usize>;
}
```

The `Component` is our only trait in the model module at the moment, aimed at storing defining traits.

## Example

A sample of how to use this can be found in the `tests` directory. 

```rust
let mut circuit_builder = CircuitBuilder::<BinaryGate>::new();

let gate1 = BinaryGate {
    inputs: vec![0, 1],
    output: 2,
};
let gate2 = BinaryGate {
    inputs: vec![3, 4],
    output: 5,
};

circuit_builder
    .add_gate(gate1)
    .add_gate(gate2);

let circuit = circuit_builder.build();
```


brech1:
- The binary module is just an example of how to use the generic structs with a basic binary circuit. So can be seen as temporary to review the pr or be left aside.
- The way we decide to deal with declaring inputs and outputs of the circuit, or this kind of "builder api" can shape the interface or even remove it completely. Current implementation is just a basic one to test the structure. In this point perhaps we could prioritize to make things easier to deal with `mpz` or adapt to the old structure as much as possible.

brech1:
The model has been updated with few key changes:

- Both circuit and gates can be modeled as:
  - components: blocks with inputs and outputs
  - executables: can perform external operations on memory
  
A key change is that the execution is defined externally on both, by part of an `Executor`.

This means that each structure can perform multiple operations with the same configuration.

sinui0 (commented):
I'm not sure I'm convinced of the execution abstractions here. For now I think it would be easiest to just stick to developing the `Circuit` type and have it enforce the invariants we discussed previously, eg topological sort etc. We can add execution in a follow up PR.

crates/mpz-circuits-generic/src/circuit.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Returns the gates.
+    pub fn gates(&self) -> &[T] {
+        &self.gates
+    }
+}
+
+/// Circuit errors.
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum CircuitError {
```
sinui0:
```suggestion
pub enum CircuitBuilderError {
```

crates/mpz-circuits-generic/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod circuit;
+pub mod model;
```
sinui0: Let's make these private and export needed types

crates/mpz-circuits-generic/tests/sorting.rs
```diff
@@ -0,0 +1,188 @@
...
+        self.inputs.clone()
+    }
+
+    fn get_outputs(&self) -> Vec<usize> {
+        vec![self.output]
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
sinui0, line 25: Can we add a couple test cases?

1. Unordered gates
2. Disconnected gate
3. Node ids are contiguous
4. Node ids are ordered: { input nodes } .. { output nodes }

crates/mpz-circuits-generic/src/model.rs
```diff
@@ -0,0 +1,12 @@
+//! Model module.
+//!
+//! This module contains the main traits and structures used to represent the circuits.
+
+/// A `Component` trait that represents a block with inputs and outputs.
+pub trait Component {
+    /// Returns the input node indices.
+    fn get_inputs(&self) -> Vec<usize>;
```
sinui0: We should return `impl Iterator<Item = &Node>` so that it's possible for the ids to be stack allocated, ie `[Node; 2]`, or some other data structure. Otherwise every call to this function is going to incur a heap allocation.

crates/mpz-circuits-generic/src/model.rs
sinui0, line 1: Can we add a `Node` newtype, as we currently have in `mpz-circuits`. Gives us better typing. Perhaps we should also switch it from `usize` to `u32` so the circuit isn't needlessly twice the size on 64 bit archs.

sinui0 (changes requested):
Nice, this will be a much more flexible base for our circuit models!

Some requests :pray:

crates/mpz-circuits-generic/src/model.rs
brech1: - I think that adding a `Node` type is not strictly necessary so I wouldn't implement it in the model. Don't see how much of an improvement this extra typing could provide on our level, perhaps implementers could add this layer themselves.
- The input or output is intended to be a position on a linear memory array so it will be `usize` in the end, on 64-bit architectures you'll have to complete the address to access an index.

But both are good suggestions, I could be missing something!

sinui0: > Don't see how much of an improvement this extra typing could provide on our level

It's all about the public API: What do you want to commit to? Using `usize` commits to that underlying representation, whereas a `Node(u32)` type allows us to constrain the data type and the traits which a node implements. For example why would a node implement `Add`? `usize` does. Do we want to allow it to be mutated? How might that affect certain invariants we have elsewhere? This is less an issue for private types, but this is explicitly part of the crates pub api.

> The input or output is intended to be a position on a linear memory array so it will be usize in the end

Yes it will be used to index into linear memory, but a `Vec<u32>` is half the size of a `Vec<u64>`, and hence a `Vec<Gate>` will be half the size if `Node` is backed by a `u32` instead of `u64` on 64-bit archs.

crates/mpz-circuits-generic/tests/sorting.rs
brech1: I added the disconnected gate test, but I'm confused about testing the node ids, in which way they should be contiguous or ordered? The order is based on that the gates should be ready to be processed, so node ids are not always contiguous or ordered

crates/mpz-circuits-generic/tests/sorting.rs
```diff
@@ -100,89 +78,135 @@ mod tests {
...
+            .add_gate(gate3)
+            .add_gate(disconnected_gate);
+
+        // Build circuit
+        let circuit = circuit_builder.build();
+        assert!(
+            circuit.is_ok(),
+            "Failed to build circuit: {:?}",
+            circuit.err()
+        );
```
sinui0, line 183: This should be an error, no? A circuit should not have disconnected gates

crates/mpz-circuits-generic/tests/sorting.rs
sinui0: > The order is based on that the gates should be ready to be processed

Exactly, so the node id should correspond to the gate position. For example:

```rust
assert!(gates.windows(2).map(|window| window[1].out.0 == window[0].out.0 + 1).all());
```

This asserts the nodes are ordered and contiguous.

crates/mpz-circuits-generic/src/model.rs
```diff
@@ -2,11 +2,24 @@
...
 
-    /// Returns the output node indices.
-    fn get_outputs(&self) -> Vec<usize>;
+    /// Returns an iterator over the output node indices.
+    fn get_outputs(&self) -> impl Iterator<Item = &Node<u32>>;
+}
+
+/// A circuit node, holds a generic type identifier.
+#[derive(Debug, Eq, PartialEq)]
+pub struct Node<T> {
```
sinui0: Why is this generic?

brech1:
Closing for #156

PR #124, opened 2024-05-08
refactor(mpz-circuits): binary type abstractions

sinui0:
This PR is an intermediate step towards making `mpz-circuits` generic over the field/elements it operates on.

Both `mpz-memory` and `mpz-binary-types` will be integrated into our VM abstractions and should remove a considerable amount of duplication.

# Changes

- Moves the dynamic typing stuff for binary values into `mpz-binary-types`
- Moves the representation abstractions into `mpz-memory`.
- Updates `mpz-circuits` to use these new crates.
  - Deprecrates the `trace` macro in `mpz-circuits` as we don't use it or intend to support it later.


sinui0:
Hold off on reviewing this, I've run into limitations with the approach that have to be addressed

PR #126, opened 2024-05-10
feat(mpz-common): add try_/join convenience macros

sinui0:
This PR adds a couple macros which makes it easier to call the `Context::join` methods without having to deal with importing the scoped future stuff or the verbosity of closures.


th4s (approved):
:+1:

themighty1 (approved):
ack

PR #132, opened 2024-05-13
feat(mpz-common): dummy executor

sinui0:
This PR adds a dummy executor that does nothing.

### DIFF
--- a/crates/mpz-common/src/executor/dummy.rs
+++ b/crates/mpz-common/src/executor/dummy.rs
@@ -1,0 +1,148 @@
+use async_trait::async_trait;
+
+use scoped_futures::ScopedBoxFuture;
+use serio::{Sink, Stream};
+
+use crate::{context::Context, ThreadId};
+
+/// A dummy executor.
+#[derive(Debug, Default)]
+pub struct DummyExecutor {
+    id: ThreadId,
+    io: DummyIo,
+}
+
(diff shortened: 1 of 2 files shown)


themighty1 (approved):

crates/mpz-common/src/executor/dummy.rs
```diff
@@ -0,0 +1,148 @@
...
+        A: for<'b> FnOnce(&'b mut Self) -> ScopedBoxFuture<'a, 'b, Result<RA, E>> + Send + 'a,
+        B: for<'b> FnOnce(&'b mut Self) -> ScopedBoxFuture<'a, 'b, Result<RB, E>> + Send + 'a,
+        RA: Send + 'a,
+        RB: Send + 'a,
+        E: Send + 'a,
+    {
+        let a = a(self).await?;
+        let b = b(self).await?;
+        Ok((a, b))
+    }
```
th4s, line 97: Shouldn't these be polled concurrently? Or doesn't it matter because it is only a dummy?

th4s (approved):

crates/mpz-common/src/executor/dummy.rs
sinui0: Doesn't really matter, probably better to do it sequentially to catch any potential deadlocks

PR #133, opened 2024-05-13
feat(mpz-common): simple counter

sinui0:
This PR adds a simple counter.

### DIFF
--- a/crates/mpz-common/src/id.rs
+++ b/crates/mpz-common/src/id.rs
@@ -53,7 +53,7 @@ impl AsRef<[u8]> for ThreadId {
 }
 
 /// A simple counter.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Counter(u32);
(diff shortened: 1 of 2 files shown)


crates/mpz-common/src/id.rs
```diff
@@ -50,3 +51,27 @@ impl AsRef<[u8]> for ThreadId {
...
+impl Counter {
+    /// Increments the counter in place, returning the previous value.
+    pub fn next(&mut self) -> Self {
+        let prev = self.0;
+        self.0 += 1;
+        Self(prev)
+    }
+
+    /// Returns the next value without incrementing the counter.
+    pub fn peek(&self) -> Self {
```
th4s, line 68: How will `peek` work? I do not see a way to inspect the inner `u32`?

th4s (approved):

crates/mpz-common/src/id.rs
sinui0: The inner type isn't part of the API, though you're right that this type is currently useless :upside_down_face: I forgot to add comparison derives

PR #135, opened 2024-05-14
Add crate `mpz-ole-core`

th4s:
This PR adds the core crate for OLE, and makes ideal random OT more generic.

### DIFF
--- a/crates/mpz-fields/src/lib.rs
+++ b/crates/mpz-fields/src/lib.rs
@@ -37,7 +37,6 @@ pub trait Field:
     + GetBit<Msb0>
     + BitLength
     + Unpin
-    + ByteRepr
 {
     /// The number of bits of a field element.
     const BIT_SIZE: u32;
@@ -71,14 +70,6 @@ pub trait Field:
 pub trait UniformRand: Sized {
     /// Return a random field element.
     fn rand<R: Rng + ?Sized>(rng: &mut R) -> Self;
-}
-
-/// The byte representation of the field.
-///
-/// This trait introduces an associated type for field elements to simplify usage.
-pub trait ByteRepr {
-    /// The underlying representation of the field element
-    type Serialized: AsRef<[u8]> + Send + Sync + Unpin + 'static;
 }
 
 impl<T> UniformRand for T
--- a/crates/mpz-fields/src/lib.rs
+++ b/crates/mpz-fields/src/lib.rs
@@ -12,6 +12,7 @@ use std::{
     ops::{Add, Mul, Neg},
 };
 
+use hybrid_array::ArraySize;
 use itybity::{BitLength, FromBitIterator, GetBit, Lsb0, Msb0};
 use rand::{distributions::Standard, prelude::Distribution, Rng};
 
@@ -35,9 +36,14 @@ pub trait Field:
     + GetBit<Lsb0>
     + GetBit<Msb0>
     + BitLength
+    + Unpin
+    + ByteRepr
 {
     /// The number of bits of a field element.
     const BIT_SIZE: u32;
+
+    /// The number of bits of a field element as a type number.
+    type BitSizeType: ArraySize;
 
     /// Return the additive identity element.
     fn zero() -> Self;
@@ -65,6 +71,14 @@ pub trait Field:
 pub trait UniformRand: Sized {
     /// Return a random field element.
     fn rand<R: Rng + ?Sized>(rng: &mut R) -> Self;
+}
+
+/// The byte representation of the field.
+///
+/// This trait introduces an associated type for field elements to simplify usage.
+pub trait ByteRepr {
+    /// The underlying representation of the field element
+    type Serialized: AsRef<[u8]> + Send + Sync + Unpin + 'static;
 }
 
 impl<T> UniformRand for T
--- a/crates/mpz-fields/src/lib.rs
+++ b/crates/mpz-fields/src/lib.rs
@@ -15,6 +15,7 @@ use std::{
 use hybrid_array::ArraySize;
 use itybity::{BitLength, FromBitIterator, GetBit, Lsb0, Msb0};
 use rand::{distributions::Standard, prelude::Distribution, Rng};
+use typenum::Unsigned;
 
 /// A trait for finite fields.
 pub trait Field:
@@ -39,10 +40,10 @@ pub trait Field:
     + Unpin
 {
     /// The number of bits of a field element.
-    const BIT_SIZE: u32;
+    const BIT_SIZE: usize = <Self::BitSize as Unsigned>::USIZE;
 
     /// The number of bits of a field element as a type number.
-    type BitSizeType: ArraySize;
+    type BitSize: ArraySize;
 
     /// Return the additive identity element.
     fn zero() -> Self;
@@ -136,11 +137,11 @@ mod tests {
     }
 
     pub(crate) fn test_field_bit_ops<T: Field>() {
-        let mut a = vec![false; T::BIT_SIZE as usize];
-        let mut b = vec![false; T::BIT_SIZE as usize];
+        let mut a = vec![false; T::BIT_SIZE];
+        let mut b = vec![false; T::BIT_SIZE];
 
         a[0] = true;
-        b[T::BIT_SIZE as usize - 1] = true;
+        b[T::BIT_SIZE - 1] = true;
 
         let a = T::from_lsb0_iter(a);
         let b = T::from_lsb0_iter(b);
@@ -148,7 +149,7 @@ mod tests {
         assert_eq!(a, T::one());
         assert!(GetBit::<Lsb0>::get_bit(&a, 0));
 
-        assert_eq!(b, T::two_pow(T::BIT_SIZE - 1));
+        assert_eq!(b, T::two_pow(T::BIT_SIZE as u32 - 1));
         assert!(GetBit::<Lsb0>::get_bit(&b, (T::BIT_SIZE - 1) as usize));
     }
 }
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -34,7 +34,8 @@ pub struct MaskedInput<const N: usize, F>(pub(crate) [F; N]);
 /// The exchange field element for share adjustment.
 ///
 /// This needs to be sent to each other in order to complete the share adjustment.
-pub struct ShareAdjust<F>(F);
+#[derive(Debug)]
+pub struct ShareAdjust<F>(pub(crate) F);
 
 #[cfg(test)]
 mod tests {
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -8,28 +8,16 @@
 mod receiver;
 mod sender;
 
+use hybrid_array::Array;
 pub use receiver::{ReceiverAdjust, ReceiverShare};
 pub use sender::{SenderAdjust, SenderShare};
 
 use mpz_fields::Field;
 
-/// Workaround because of feature `generic_const_exprs` not available in stable.
-///
-/// This is used to check at compile-time that the correct const-generic implementation is used for
-/// a specific field.
-struct Check<const N: usize, F: Field>(std::marker::PhantomData<F>);
-
-impl<const N: usize, F: Field> Check<N, F> {
-    const IS_BITSIZE_CORRECT: () = assert!(
-        N as u32 == F::BIT_SIZE,
-        "Wrong bit size used for field. You need to use `F::BIT_SIZE` for N."
-    );
-}
-
 /// The masked input of the sender.
 ///
 /// This is the correlation which is sent to the receiver and hides the sender's input.
-pub struct MaskedInput<const N: usize, F>(pub(crate) [F; N]);
+pub struct MaskedInput<F: Field>(pub(crate) Array<F, F::BitSizeType>);
 
 /// The exchange field element for share adjustment.
 ///
@@ -74,7 +62,7 @@ mod tests {
         let (ot_messages, ot_message_choices) = create_rot(receiver_input.clone());
 
         let (sender_shares, masked) =
-            SenderShare::new_vec::<256>(sender_input.clone(), ot_messages).unwrap();
+            SenderShare::new_vec(sender_input.clone(), ot_messages).unwrap();
         let receiver_shares =
             ReceiverShare::new_vec(receiver_input.clone(), ot_message_choices, masked).unwrap();
 
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -4,6 +4,7 @@
 //! - The `Initialize` stage is instantiated using random OT rather than chosen-input OT.
 //! - The `Extend` stage can only be called once, since our goal is to implement oblivious linear
 //!   function evaluation (OLE) rather than vector OLE (VOLE) (which means that we do not use PRGs).                                                  
+//! - The evaluated function is f(b)=a*b+x rather than f(b)=a*b-x.                                                  
 //!                                                                                       
 //! Note that this is an OLE with errors implementation. A malicious sender is allowed to set its own
 //! output and can introduce additive errors into the receiver's output.
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -20,7 +20,7 @@ use mpz_fields::Field;
 /// The masked correlation of the sender.
 ///
 /// This is the correlation which is sent to the receiver.
-pub struct MaskedCorrelation<F: Field>(pub(crate) Array<F, F::BitSizeType>);
+pub struct MaskedCorrelation<F: Field>(pub(crate) Array<F, F::BitSize>);
 
 /// The exchange field element for share adjustment.
 ///
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -88,7 +88,6 @@ mod tests {
         assert_eq!(y, a * b + x);
     }
 
-    // N should be BIT_SIZE of F
     fn create_ole(
         sender_input: P256,
         receiver_input: P256,
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -17,10 +17,10 @@ pub use sender::{SenderAdjust, SenderShare};
 
 use mpz_fields::Field;
 
-/// The masked input of the sender.
+/// The masked correlation of the sender.
 ///
-/// This is the correlation which is sent to the receiver and hides the sender's input.
-pub struct MaskedInput<F: Field>(pub(crate) Array<F, F::BitSizeType>);
+/// This is the correlation which is sent to the receiver.
+pub struct MaskedCorrelation<F: Field>(pub(crate) Array<F, F::BitSizeType>);
 
 /// The exchange field element for share adjustment.
 ///
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -4,13 +4,12 @@
 //! vector OLE (VOLE), which means that we do not use PRGs, i.e. Extend can only be called once.                                                  
 //!                                                                                       
 //! Note that this is an OLE with errors implementation. The sender can introduce additive errors.
-//! Input privacy is guaranteed, but output privacy is not, when `0` is chosen as an input value.
 
 mod receiver;
 mod sender;
 
-pub(crate) use receiver::{ReceiverAdjust, ReceiverShare};
-pub(crate) use sender::{SenderAdjust, SenderShare};
+pub use receiver::{ReceiverAdjust, ReceiverShare};
+pub use sender::{SenderAdjust, SenderShare};
 
 use mpz_fields::Field;
 
@@ -30,7 +29,7 @@ impl<const N: usize, F: Field> Check<N, F> {
 /// The masked input of the sender.
 ///
 /// This is the correlation which is sent to the receiver and hides the sender's input.
-pub struct MaskedInput<const N: usize, F>([F; N]);
+pub struct MaskedInput<const N: usize, F>(pub(crate) [F; N]);
 
 /// The exchange field element for share adjustment.
 ///
@@ -39,7 +38,7 @@ pub struct ShareAdjust<F>(F);
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use crate::core::{ReceiverShare, SenderShare};
     use crate::tests::create_rot;
     use mpz_core::{prg::Prg, Block};
     use mpz_fields::{p256::P256, UniformRand};
@@ -60,6 +59,30 @@ mod tests {
         let y = receiver_share.inner();
 
         assert_eq!(y, a * b + x);
+    }
+
+    #[test]
+    fn test_ole_core_vec() {
+        let count = 12;
+        let from_seed = Prg::from_seed(Block::ZERO);
+        let mut rng = from_seed;
+
+        let sender_input: Vec<P256> = (0..count).map(|_| P256::rand(&mut rng)).collect();
+        let receiver_input: Vec<P256> = (0..count).map(|_| P256::rand(&mut rng)).collect();
+
+        let (ot_messages, ot_message_choices) = create_rot(receiver_input.clone());
+
+        let (sender_shares, masked) =
+            SenderShare::new_vec::<256>(sender_input.clone(), ot_messages).unwrap();
+        let receiver_shares =
+            ReceiverShare::new_vec(receiver_input.clone(), ot_message_choices, masked).unwrap();
+
+        sender_input
+            .iter()
+            .zip(receiver_input)
+            .zip(sender_shares)
+            .zip(receiver_shares)
+            .for_each(|(((&a, b), x), y)| assert_eq!(y.inner(), a * b + x.inner()));
     }
 
     #[test]
--- a/crates/mpz-ole-core/src/core/mod.rs
+++ b/crates/mpz-ole-core/src/core/mod.rs
@@ -1,0 +1,113 @@
+//! This implementation uses the COPEe protocol from <https://eprint.iacr.org/2016/505> page 10.
+//!
+//! We use this construction to implement oblivious linear function evaluation (OLE) instead of
+//! vector OLE (VOLE), which means that we do not use PRGs, i.e. Extend can only be called once.                                                  
+//!                                                                                       
+//! Note that this is an OLE with errors implementation. The sender can introduce additive errors.
+//! Input privacy is guaranteed, but output privacy is not, when `0` is chosen as an input value.
+
+mod receiver;
+mod sender;
+
+pub(crate) use receiver::{ReceiverAdjust, ReceiverShare};
+pub(crate) use sender::{SenderAdjust, SenderShare};
+
+use mpz_fields::Field;
+
+/// Workaround because of feature `generic_const_exprs` not available in stable.
+///
+/// This is used to check at compile-time that the correct const-generic implementation is used for
+/// a specific field.
+struct Check<const N: usize, F: Field>(std::marker::PhantomData<F>);
+
+impl<const N: usize, F: Field> Check<N, F> {
+    const IS_BITSIZE_CORRECT: () = assert!(
+        N as u32 == F::BIT_SIZE,
+        "Wrong bit size used for field. You need to use `F::BIT_SIZE` for N."
+    );
+}
+
+/// The masked input of the sender.
+///
+/// This is the correlation which is sent to the receiver and hides the sender's input.
+pub struct MaskedInput<const N: usize, F>([F; N]);
+
+/// The exchange field element for share adjustment.
+///
+/// This needs to be sent to each other in order to complete the share adjustment.
+pub struct ShareAdjust<F>(F);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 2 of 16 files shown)


crates/mpz-ole-core/src/core/mod.rs
```diff
@@ -0,0 +1,108 @@
+//! This implementation uses the COPEe protocol from <https://eprint.iacr.org/2016/505> page 10.
+//!
+//! We use this construction to implement oblivious linear function evaluation (OLE) instead of
+//! vector OLE (VOLE), which means that we do not use PRGs, i.e. Extend can only be called once.                                                  
+//!                                                                                       
+//! Note that this is an OLE with errors implementation. The sender can introduce additive errors.
+//! Input privacy is guaranteed, but output privacy is not, when `0` is chosen as an input value.
```
sinui0: I think we can remove this line

crates/mpz-ole-core/src/core/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+
+mod receiver;
+mod sender;
+
+pub(crate) use receiver::{ReceiverAdjust, ReceiverShare};
+pub(crate) use sender::{SenderAdjust, SenderShare};
```
sinui0: These can be `pub`

crates/mpz-ole-core/src/core/receiver.rs
```diff
@@ -0,0 +1,92 @@
+//! Receiver shares for Oblivious Linear Function Evaluation (OLE).
+
+use super::{Check, MaskedInput, ShareAdjust};
```
sinui0: Please use `use crate:: ..`, `super` makes refactoring more burdensome

crates/mpz-ole-core/src/ideal.rs
```diff
@@ -0,0 +1,72 @@
...
+            sender_input.len(),
+            receiver_input.len(),
+            "Vectors of field elements should have equal length."
+        );
+
+        let sender_output: Vec<F> = (0..sender_input.len())
+            .map(|_| F::rand(&mut self.0))
+            .collect();
+
+        let reciever_output: Vec<F> = sender_input
```
sinui0: typo

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum OLEError {
+    #[error("The number of OTs is incorrect. Got {0}, expected {1}")]
+    ExpectedMultipleOf(usize, usize),
+    #[error("Not enough prepared OLEs available. Requested {0}, but only {1} are available")]
+    InsufficientOLEs(usize, usize),
+    #[error("Number of adjustments has to be equal. Got {0} and {1}")]
+    UnequalAdjustments(usize, usize),
+    #[error("Provided number of masks is incorrect. Got {0}, expected {1}")]
+    WrongNumerOfMasks(usize, usize),
```
sinui0:
```suggestion
    WrongNumberOfMasks(usize, usize),
```

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+use mpz_fields::Field;
+
+use crate::{
+    core::{MaskedInput, ReceiverAdjust, ReceiverShare, ShareAdjust},
+    OLEError,
+};
+
+/// A receiver for batched OLE.
+#[derive(Debug)]
+pub struct OLEReceiver<const N: usize, F> {
```
sinui0: does this even compile? `N` isn't used anywhere

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl<const N: usize, F: Field> OLEReceiver<N, F> {
+    /// Generates new OLEs.
+    ///
+    /// OLEs are not stored and directly returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The receiver's OLE input factors.
+    /// * `ot_message_choices` - The OT messages chosen by the receiver.
```
sinui0: We shouldn't refer to these as "OT messages", that is an unnecessary implementation detail. They just need to be uniformly random field elements.

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// * `masked` - The correlations from the sender.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`ReceiverShare`]s containing the OLE outputs for the receiver.
+    pub fn generate(
```
sinui0: This method can just be an associated method on `ReceiverShare`, eg `ReceiverShare::new_vec`

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// For consumption of OLEs which have been stored by [`OLEReceiver::preprocess`].
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of shares to return.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`ReceiverShare`]s containing the OLE outputs for the receiver.
+    pub fn consume(&mut self, count: usize) -> Result<Vec<ReceiverShare<F>>, OLEError> {
```
sinui0: Return type can be `Option<Vec<_>>` instead of error.

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `targets` - The new OLE receiver inputs.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`ReceiverAdjust`] which needs to be converted by [`OLEReceiver::finish_adjust`].
+    /// * A vector of [`ShareAdjust`] which needs to be sent to the [`super::OLESender`].
+    pub fn adjust(
```
sinui0: This should store the intermediate values in a separate buffer and return a message type to send the sender, eg:

```rust
struct BatchAdjust<F> {
  adjustments: Vec<ShareAdjust<F>>
}
```

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// * `receiver_adjust` - The receiver's intermediate shares from [`OLEReceiver::adjust`].
+    /// * `adjustments` - The sender's adjustments.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`ReceiverShare`]s containing the new OLE outputs for the receiver.
+    pub fn finish_adjust(
+        &self,
+        receiver_adjust: Vec<ReceiverAdjust<F>>,
+        adjustments: Vec<ShareAdjust<F>>,
```
sinui0: This should only accept `BatchAdjust<F>` and then apply them to the new buffer we added above.

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// # Arguments
+    ///
+    /// * `input` - The sender's OLE input factors.
+    /// * `ot_messages` - The OT messages used for generating correlations.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`SenderShare`]s containing the OLE outputs for the sender.
+    /// * A vector of [`MaskedInput`]s, which is to be sent to the receiver.
+    pub fn generate(
```
sinui0: Same here, associated method on `SenderShare`

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// * `ot_messages` - The OT messages used for generating correlations.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`MaskedInput`]s, which is to be sent to the [`super::OLEReceiver`].
+    pub fn preprocess(
+        &mut self,
+        input: Vec<F>,
+        ot_messages: Vec<[F; 2]>,
+    ) -> Result<Vec<MaskedInput<N, F>>, OLEError> {
```
sinui0: Define a message type, eg `MaskedInputs<F>`

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// For consumption of OLEs which have been stored by [`OLESender::preprocess`].
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of shares to return.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`SenderShare`]s containing the OLE output for the sender.
+    pub fn consume(&mut self, count: usize) -> Result<Vec<SenderShare<F>>, OLEError> {
```
sinui0: Return `Option`

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// * `targets` - The new OLE sender inputs.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`SenderAdjust`]s which needs to be converted by [`OLESender::finish_adjust`].
+    /// * A vector of [`ShareAdjust`]s which needs to be sent to the [`super::OLEReceiver`].
+    pub fn adjust(
+        &mut self,
+        targets: Vec<F>,
+    ) -> Result<(Vec<SenderAdjust<F>>, Vec<ShareAdjust<F>>), OLEError> {
```
sinui0: Return `BatchAdjust<F>`, store intermediates

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// * `sender_adjust` - The sender's intermediate shares from [`OLESender::adjust`].
+    /// * `adjustments` - The receiver's adjustments.
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`SenderShare`]s containing the new OLE outputs for the sender.
+    pub fn finish_adjust(
+        &self,
+        sender_adjust: Vec<SenderAdjust<F>>,
+        adjustments: Vec<ShareAdjust<F>>,
```
sinui0: Receive `BatchAdjust<F>`

crates/mpz-ot-core/src/ideal/rot.rs
sinui0, line 1: :+1:

sinui0 (changes requested):

crates/mpz-ole-core/src/receiver.rs
th4s: It is used in `preprocess`. Const Generics do not require `PhantomData`. https://github.com/rust-lang/rust/issues/83330

crates/mpz-ole-core/src/receiver.rs
th4s: Hmm what should happen if I request 10 OLEs but only 5 are available?

crates/mpz-ole-core/src/sender.rs
th4s: same as above

crates/mpz-ole-core/src/receiver.rs
th4s: This would mean that we have to introduce some `HashMap` and ticketing system, right? Because otherwise consumers of this API have to call `adjust` and `finish_adjust` in the right order. So for example if you have consumer `A` and `B` and they do it in this order, then it would **not** work.
`A` -> `adjust`, `B` -> `adjust`, `B` -> `finish_adjust`, `A` -> `finish_adjust`

That's why I chose to return it instead of storing it.

crates/mpz-ole-core/src/receiver.rs
sinui0: There is only 1 error condition: whether or not the OLEs exist. You would return `None` in your example

crates/mpz-ole-core/src/receiver.rs
sinui0: I think requiring that it is done in order is the desired behavior. To detect when it is not done in order, let's add a `TransferId` the same as we do in the OT crates. These types offer very little otherwise.

th4s:
Ok I think you can take a look again. I did it similar to `mpz-ot-core`.

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,143 @@
...
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`ReceiverShare`]s containing the OLE outputs for the receiver.
+    pub fn consume(&mut self, count: usize) -> Option<Vec<ReceiverShare<F>>> {
+        if count > self.cache.len() {
+            return None;
+        }
+
+        let shares = self.cache.drain(..count).collect();
```
sinui0, line 64: Let's use [`Vec::split_off`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.split_off) here

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    /// # Returns
+    ///
+    /// * A vector of [`SenderShare`]s containing the OLE output for the sender.
+    pub fn consume(&mut self, count: usize) -> Option<Vec<SenderShare<F>>> {
+        if count > self.cache.len() {
+            return None;
+        }
+
+        let shares = self.cache.drain(..count).collect();
```
sinui0, line 65: And here

sinui0 (approved):
Good work, lgtm.

Only 1 nit

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -61,7 +61,7 @@ impl<const N: usize, F: Field> OLEReceiver<N, F> {
             return None;
         }
 
-        let shares = self.cache.drain(..count).collect();
+        let shares = self.cache.split_off(count);
```
sinui0: This splits off the end, not the front

th4s: Shouldn't matter that we take the end instead of the front. But the number of OLEs this returns is wrong, right?

sinui0: The expected behavior would be FIFO. You can achieve this with `mem::swap` to put the tail back and return the head

sinui0: Or VecDeque

th4s: Hmm, why do you think that the expected behavior is FIFO? My understanding is that the API makes no such guarantees.

- The thing with `mem::swap` is that you then return the full-capacity `Vec`, so further calls to `preprocess` would have to reallocate more often and the big chunk of the `Vec` which remains in `self` has to be reallocated for every call of `consume`. This effect gets worse, the smaller the part is that you `split_off` and `swap`.
- Which method from `VecDeque` do you mean, `rotate_right`  and then `split_off`? I think this is more inefficient than a `split_off` on `Vec`.

What advantages do you see in guaranteeing FIFO access?

sinui0: I would just switch back to `drain` on `VecDeque`, its front drain implementation is probably more efficient.

> What advantages do you see in guaranteeing FIFO access?

FIFO is not surprising behavior, other orderings are

th4s: You are right, I benched it.
Screenshot_2024-05-21_17-53-52

crates/mpz-ole-core/src/core/sender.rs
```diff
@@ -0,0 +1,133 @@
...
+        // Check that the right N is used depending on the needed bit size of the field.
+        let _: () = Check::<N, F>::IS_BITSIZE_CORRECT;
+
+        let output = random
+            .iter()
+            .enumerate()
+            .fold(F::zero(), |acc, (i, &[zero, _])| {
+                acc + F::two_pow(i as u32) * zero
+            });
+        let share = Self { input, output };
```
themighty1, line 42: the output share should be `-output` acc to the paper

crates/mpz-ole-core/src/core/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    const IS_BITSIZE_CORRECT: () = assert!(
+        N as u32 == F::BIT_SIZE,
+        "Wrong bit size used for field. You need to use `F::BIT_SIZE` for N."
+    );
+}
+
+/// The masked input of the sender.
+///
+/// This is the correlation which is sent to the receiver and hides the sender's input.
+pub struct MaskedInput<const N: usize, F>(pub(crate) [F; N]);
```
themighty1: The KOS16 paper calls this "The masked correlation", so it views it as if the input was used to mask the coorelation, not the other way around.
Can we stick to the paper's mental model? it will be easier to compare this protocol to the paper this way.

crates/mpz-ole-core/src/core/sender.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        Ok((shares, masked))
+    }
+
+    /// Returns the sender's output share.
+    pub fn inner(self) -> F {
+        self.output
+    }
+
+    /// Adjust a preprocessed share.
```
themighty1:
```suggestion
    /// Adjusts a preprocessed share.
```

crates/mpz-ole-core/src/core/receiver.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        Ok(shares)
+    }
+
+    /// Returns the receiver's output share.
+    pub fn inner(self) -> F {
+        self.output
+    }
+
+    /// Adjust a preprocessed share.
```
themighty1:
```suggestion
    /// Adjusts a preprocessed share.
```

crates/mpz-ole-core/src/core/sender.rs
```diff
@@ -0,0 +1,133 @@
...
+    old_output: F,
+    new_input: F,
+}
+
+impl<F: Field> SenderAdjust<F> {
+    /// Finishes the adjustment and returns the adjusted sender's share.
+    pub(crate) fn finish(self, adjust: ShareAdjust<F>) -> SenderShare<F> {
+        SenderShare {
+            input: self.new_input,
+            output: self.old_output + self.old_input * adjust.0,
```
themighty1, line 133: there must be a minus here acc to the SE post: `a′−x′v`

crates/mpz-ole-core/src/ideal.rs
```diff
@@ -0,0 +1,72 @@
+//! Ideal functionality for Oblivious Linear Function Evaluation (OLE).
+
+use mpz_fields::Field;
+use rand::{rngs::ThreadRng, thread_rng};
+
+/// The OLE functionality
```
themighty1:
```suggestion
/// The OLE functionality.
```

crates/mpz-ole-core/src/ideal.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let sender_output: Vec<F> = (0..sender_input.len())
+            .map(|_| F::rand(&mut self.0))
+            .collect();
+
+        let receiver_output: Vec<F> = sender_input
+            .iter()
+            .zip(receiver_input)
+            .zip(sender_output.iter().copied())
+            .map(|((&a, &b), x)| a * b + x)
```
themighty1, line 35: this should be a*b - x

crates/mpz-ole-core/src/core/mod.rs
```diff
@@ -0,0 +1,132 @@
+//! This implementation uses the COPEe protocol from <https://eprint.iacr.org/2016/505> page 10.
+//!
+//! We use this construction to implement oblivious linear function evaluation (OLE) instead of
+//! vector OLE (VOLE), which means that we do not use PRGs, i.e. Extend can only be called once.                                                  
+//!                                                                                       
+//! Note that this is an OLE with errors implementation. The sender can introduce additive errors.
```
themighty1: I made the description a little more verbose

```suggestion
//! This implementation is based on the COPEe protocol from <https://eprint.iacr.org/2016/505> page 10
//! with the following modification:
//!
//! - The `Initialize` stage is instantiated using random OT rather than chosen-input OT.
//! - The `Extend` stage can only be called once, since our goal is to implement oblivious linear
//!   function evaluation (OLE) rather than vector OLE (VOLE) (which means that we do not use PRGs).                                                  
//!                                                                                       
//! Note that this is an OLE with errors implementation. A malicious sender is allowed to set its own
//! output and can introduce additive errors into the receiver's output.
```

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    use mpz_core::{prg::Prg, Block};
+    use mpz_fields::{p256::P256, UniformRand};
+    use mpz_ot_core::ideal::rot::IdealROT;
+    use rand::SeedableRng;
+
+    #[test]
+    fn test_ole_sender_receiver_preprocess() {
+        let count = 12;
+        let from_seed = Prg::from_seed(Block::ZERO);
+        let mut rng = from_seed;
```
themighty1:
```suggestion
        let mut rng = Prg::from_seed(Block::ZERO);
```

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let sender_shares = sender.consume(count).unwrap();
+        let receiver_shares = receiver.consume(count).unwrap();
+
+        sender_input
+            .iter()
+            .zip(receiver_input)
+            .zip(sender_shares)
+            .zip(receiver_shares)
+            .for_each(|(((&a, b), x), y)| assert_eq!(y.inner(), a * b + x.inner()));
```
themighty1, line 96: should be a*b - x

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    #[test]
+    fn test_ole_sender_receiver_adjust() {
+        let count = 12;
+        let from_seed = Prg::from_seed(Block::ZERO);
+        let mut rng = from_seed;
```
themighty1:
```suggestion
        let mut rng = Prg::from_seed(Block::ZERO);
```

crates/mpz-ole-core/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let sender_shares_adjusted = sender_adjust.finish_adjust(r_to_s_adjust).unwrap();
+        let receiver_shares_adjusted = receiver_adjust.finish_adjust(s_to_r_adjust).unwrap();
+
+        sender_targets
+            .iter()
+            .zip(receiver_targets)
+            .zip(sender_shares_adjusted)
+            .zip(receiver_shares_adjusted)
+            .for_each(|(((&a, b), x), y)| assert_eq!(y.inner(), a * b + x.inner()));
```
themighty1, line 133: should be a*b - x

crates/mpz-ole-core/src/msg.rs
```diff
@@ -0,0 +1,45 @@
+//! Message types for OLE.
+
+use crate::{core::MaskedInput, OLEError, TransferId};
+use mpz_fields::Field;
+use serde::{Deserialize, Serialize};
+
+/// Message type for sending a vector [`MaskedInput`]s to the receiver.
```
themighty1:
```suggestion
/// Message type for sending a vector of [`MaskedInput`]s to the receiver.
```

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Adjusts OLEs in the internal cache.
+    ///
+    /// # Arguments
+    ///
+    /// * `targets` - The new OLE receiver inputs.
+    ///
+    /// # Returns
+    ///
+    /// * [`BatchReceiverAdjust`] which needs to be converted by [`BatchReceiverAdjust::finish_adjust`].
+    /// * [`BatchAdjust`] which needs to be sent to the [`crate::OLESender`].
```
themighty1: crate::OLESender was called in earlier comments simply "sender"
```suggestion
    /// * [`BatchAdjust`] which needs to be sent to the sender.
```

crates/mpz-ole-core/src/receiver.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+}
+
+impl<const N: usize, F: Field> OLEReceiver<N, F> {
+    /// Generates new OLEs and stores them internally.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The receiver's OLE input factors.
```
themighty1: Let's call them shares like in the core module

```suggestion
    /// * `input` - The receiver's OLE input shares.
```

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+
+impl<const N: usize, F: Field> OLESender<N, F> {
+    /// Generates new OLEs and stores them internally.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The sender's OLE input factors.
```
themighty1:
```suggestion
    /// * `input` - The sender's OLE input shares.
```

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// * `random` - Uniformly random field elements for the correlation.
+    ///
+    /// # Returns
+    ///
+    /// * [`MaskedInputs`], which are to be sent to the [`crate::OLEReceiver`].
```
themighty1:
```suggestion
    /// * [`MaskedInputs`], which are to be sent to the receiver.
```

crates/mpz-ole-core/src/sender.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Adjusts OLEs in the internal cache.
+    ///
+    /// # Arguments
+    ///
+    /// * `targets` - The new OLE sender inputs.
+    ///
+    /// # Returns
+    ///
+    /// * [`BatchSenderAdjust`] which needs to be converted by [`BatchSenderAdjust::finish_adjust`].
+    /// * [`BatchAdjust`] which needs to be sent to the [`crate::OLEReceiver`].
```
themighty1:
```suggestion
    /// * [`BatchAdjust`] which needs to be sent to the receiver.
```

themighty1 (changes requested):
gw! There are some comments, potentially blocking.

Since COPEe's computational security parameter depends on the Field size, can we also add a check that CSP >= Field::BIT_SIZE ?

th4s:
Thanks! Let's address all of these minus sign comments in one batch.

It all boils down to the question, if we implement COPE from the MASCOT paper as is, or if we build upon COPE to implement OLE. The difference is that COPE is `y + x = a * b` and OLE is `y = a * b + x`.

Since OLE is the more modern and widely used functionality, I decided for the latter. Can we agree to proceed that way?

crates/mpz-fields/src/lib.rs
```diff
@@ -35,10 +36,15 @@ pub trait Field:
...
     + GetBit<Msb0>
     + BitLength
+    + Unpin
+    + ByteRepr
 {
     /// The number of bits of a field element.
     const BIT_SIZE: u32;
 
+    /// The number of bits of a field element as a type number.
+    type BitSizeType: ArraySize;
```
sinui0: I think we can just call this `BitSize`

crates/mpz-ole-core/src/core/sender.rs
```diff
@@ -25,20 +26,24 @@ impl<F: Field> SenderShare<F> {
...
             });
         let share = Self { input, output };
 
-        let mut ui = [F::zero(); N];
-        ui.iter_mut()
+        let mut ui: Array<F, F::BitSizeType> = Array::from_fn(|_| F::zero());
+        ui.as_mut_slice()
+            .iter_mut()
             .zip(random)
             .for_each(|(u, [zero, one])| *u = zero + -one + input);
```
sinui0: Should be able to put this all into one expression

```rust
let ui = Array::from_fn(|i| {
 let [zero, one] = random[i];
 zero + -one + input
});
```

crates/mpz-fields/src/lib.rs
```diff
@@ -35,10 +36,15 @@ pub trait Field:
     + GetBit<Lsb0>
     + GetBit<Msb0>
     + BitLength
+    + Unpin
+    + ByteRepr
 {
     /// The number of bits of a field element.
     const BIT_SIZE: u32;
```
sinui0: This probably isn't needed anymore as `ArraySize` is a super trait of `typenum::Unsigned` so you can:

```rust
<F::BitSize as Unsigned>::USIZE
```

but if it takes more than a minute to change that I'd leave it for now

sinui0 (approved):

crates/mpz-fields/src/lib.rs
th4s: Why not add it as a default constant
```rust
    const BIT_SIZE: usize = <Self::BitSizeType as Unsigned>::USIZE;
```

sinui0: Even better, didn't realize this was possible

crates/mpz-ole-core/src/core/mod.rs
```diff
@@ -0,0 +1,123 @@
+//! This implementation is based on the COPEe protocol from <https://eprint.iacr.org/2016/505> page 10
+//! with the following modification:
+//!
+//! - The `Initialize` stage is instantiated using random OT rather than chosen-input OT.
+//! - The `Extend` stage can only be called once, since our goal is to implement oblivious linear
+//!   function evaluation (OLE) rather than vector OLE (VOLE) (which means that we do not use PRGs).                                                  
+//!
```
themighty1, line 8:
```suggestion
//! - The evaluated function is f(b)=a*b+x rather than f(b)=a*b-x.                                                  
//!                                                                                       
```

themighty1 (approved):
ack, i agree it makes sense to take the a*b+x approach

PR #136, opened 2024-05-15
feat(mpz-common): multi-threaded executor

sinui0:
This PR implements a multi-threaded executor which utilizes a multiplexed connection to support forking.

~~It is currently limited to executing 2 futures concurrently, but we can expand this as needed.~~

EDIT: I've gone all the way and added full forking support which can be accessed via the `queue` method on `Context`. This can be used to distribute an arbitrary number of tasks across a configurable number of logical threads.


crates/mpz-common/src/executor/mt.rs
```diff
@@ -0,0 +1,213 @@
...
+    M: FramedUidMux<ThreadId, Framed = Io> + Clone,
+    M::Error: Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
+{
+    async fn fork(&mut self) -> Result<Box<Self>, ContextError> {
+        // Forking a thread context is only performed once, afterwhich the child ctx
+        // is stored for later use.
+
+        if let Some(child) = self.child.take() {
+            return Ok(child);
+        }
```
th4s: Shouldn't this be an error, because the caller does not know if he created a new thread or gets an existing one.

EDIT: I realize this method is not public, so I think it should be ok.

th4s (approved):
Very nice :+1:

crates/mpz-common/src/queue.rs
```diff
@@ -0,0 +1,32 @@
...
+/// An async task which is executed with a mutable reference to state.
+pub trait Task<S: ?Sized, R>: for<'a> FnOnce(&'a mut S) -> TaskFut<'a, R> + Send + 'static {}
+
+impl<S, R, T> Task<S, R> for T where T: for<'a> FnOnce(&'a mut S) -> TaskFut<'a, R> + Send + 'static {}
+
+/// A future that yields the output of a task.
+pub type TaskFut<'a, R> = Pin<Box<dyn Future<Output = R> + Send + 'a>>;
+
+/// A queue of tasks which can be executed concurrently with access to shared state.
+pub trait Queue<S: ?Sized, R> {
```
th4s: All Futures must resolve to the same return type, right? Is that wanted behavior?

crates/mpz-common/src/queue/round_robin.rs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        self.queue.push_back(Box::new(f));
+    }
+
+    fn wait(&mut self) -> impl Future<Output = Result<Vec<R>, ContextError>> + Send {
+        let lane_count = self.lanes.len();
+        let task_count = self.queue.len();
+
+        let mut lanes: Vec<_> = (0..lane_count)
+            .map(|_| Vec::with_capacity((task_count / lane_count) + (task_count % lane_count)))
```
th4s:
```suggestion
            .map(|_| Vec::with_capacity((task_count / lane_count) + 1))
```
One more should be enough, right?

crates/mpz-common/src/queue/round_robin.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        // Create futures which process each lane with the corresponding state.
+        let lane_futs: Vec<_> = lanes
+            .into_iter()
+            .zip(self.lanes.iter_mut())
+            .map(|(lane, ctx)| async move {
+                let mut results = Vec::with_capacity(lane.len());
+                for task in lane {
+                    results.push(task(ctx).await);
```
th4s: The call to `await` here, will drive a single future to completion here, right? We should collect them without calling `await` so that they can be polled concurrently.

EDIT: I just saw the `async` block wrapper. Then it should be ok since the wrapper future will be lazily evaluated?

th4s (commented):
Nice :+1:, some comments

crates/mpz-common/src/executor/mt.rs
```diff
@@ -0,0 +1,358 @@
...
+                ctx.io_mut().send(1u8).await.unwrap();
+            })
+        });
+        queue_b.push(|ctx| Box::pin(async { ctx.io_mut().expect_next::<u8>().await.unwrap() }));
+
+        let (_, results_b) = futures::try_join!(queue_a.wait(), queue_b.wait()).unwrap();
+
+        assert_eq!(results_b, vec![0, 1]);
+    }
+}
```
themighty1, line 295: Can we add a test to make sure things run concurrently

```rust


    use std::time::{Duration, Instant};
    // Need to enable tokio time feature

    // Tests that the tasks run concurrently.
    #[tokio::test]
    async fn test_mt_executor_concurrent() {
        let ((mux_a, fut_a), (mux_b, fut_b)) = test_yamux_pair_framed(1024, Bincode);

        tokio::spawn(async move {
            futures::try_join!(fut_a.into_future(), fut_b.into_future()).unwrap();
        });

        let mut exec_a = MTExecutor::new(mux_a, 8);
        let mut exec_b = MTExecutor::new(mux_b, 8);

        let (mut ctx_a, mut ctx_b) =
            futures::try_join!(exec_a.new_thread(), exec_b.new_thread()).unwrap();

        let mut queue_a = ctx_a.queue().await.unwrap();
        let mut queue_b = ctx_b.queue().await.unwrap();

        let start = Instant::now();
        let timeout = Duration::from_millis(100);

        for i in 0..8 {
            queue_a.push(move |ctx| {
                let i = i;
                let timeout = timeout;
                Box::pin(async move {
                    tokio::time::sleep(timeout).await;
                    ctx.io_mut().send(i as u8).await.unwrap();
                })
            });
            queue_b.push(|ctx| Box::pin(async { ctx.io_mut().expect_next::<u8>().await.unwrap() }));
        }

        let (_, results_b) = futures::try_join!(queue_a.wait(), queue_b.wait()).unwrap();

        let elapsed = Instant::now().duration_since(start);

        // The overall latency should be approximately that of a single task.
        assert!(elapsed < timeout + Duration::from_millis(50));

        assert_eq!(results_b, vec![0, 1, 2, 3, 4, 5, 6, 7]);
    }
```

crates/mpz-common/src/queue.rs
sinui0: Yeah, this almost mirrors `FuturesOrdered`

crates/mpz-common/src/queue/round_robin.rs
sinui0: Each task in a lane is processed one at a time, as `&mut ctx` is passed into the task. The lanes are processed concurrently

crates/mpz-common/src/queue/round_robin.rs
themighty1: should be + (task_count % lane_count != 0)

themighty1 (approved):
gw

sinui0:
I'm going to just remove this queue stuff and stick with the other changes. Will proceed with merge as scope will be a subset. Will open another PR if/when we need this functionality

sinui0:
Sorry for changing scope after review. I also identified a bug which I've fixed.

crates/mpz-common/src/executor/mt.rs
```diff
@@ -141,11 +119,16 @@ where
             .take()
             .expect("children were not left uninitialized");
 
-        if children.len() < 1 {
-            children.alloc(&self.mux, 1).await?;
+        if children.len() < 2 {
```
th4s, line 122: So if `children.len() == 1`, then we would have 3 after successful allocation. But I assume this is ok?

th4s (approved):

crates/mpz-common/src/executor/mt.rs
themighty1: No, `alloc(2)` brings the total count to 2. If there already was 1 thread, it will allocate only 1 extra thread in order to bring the total to 2.

themighty1 (approved):

PR #138, opened 2024-05-20
Add IO wrapper for OLE

th4s:
This PR adds the async wrapper crate for OLE.


crates/mpz-ole/src/msg.rs
sinui0, line 1: This wrapper type should not be needed unless you want the protocol to be able to branch, which it doesn't in this case

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    OT(#[from] OTError),
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error(transparent)]
+    OLECoreError(#[from] OLECoreError),
+    #[error("Not enough OLEs available")]
+    InsufficientOLEs,
+    #[error(transparent)]
+    Message(Box<dyn Error + Send + 'static>),
+}
```
sinui0: Over time I've become "opaque error" pilled. Virtually all errors in `mpz` should just be opaque errors which do not allow the user to inspect them and try to recover. On a case by case basis we can decide to allow that if the need arises. Otherwise, their `Debug` implementation can provide debugging info in addition to logging errors via `tracing`

Something like:

```rust
#[error("ole error: {kind}")]
pub struct OleError {
  kind: OleErrorKind,
  #[source]
  source: Option<Box<dyn Error + Send + Sync>>
}

// Private enum
pub(crate) enum OleErrorKind {
  Io,
  Core,
  ..
}
```

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,92 @@
...
+use serde::{Deserialize, Serialize};
+use serio::stream::IoStreamExt;
+use serio::SinkExt;
+use std::marker::PhantomData;
+
+/// OLE receiver.
+pub struct OLEReceiver<const M: usize, const N: usize, T, F, C> {
+    rot_receiver: T,
+    core: OLECoreReceiver<N, F>,
+    context: PhantomData<C>,
```
sinui0: The struct shouldn't be coupled to a context

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,92 @@
...
+            context: PhantomData,
+        }
+    }
+
+    /// Preprocesses OLEs.
+    ///
+    /// # Arguments
+    ///
+    /// * `count` - The number of OLEs to preprocess.
+    pub async fn preprocess(&mut self, ctx: &mut C, count: usize) -> Result<(), OLEError> {
```
sinui0:
```suggestion
    pub async fn preprocess<Ctx: Context>(&mut self, ctx: &mut Ctx, count: usize) -> Result<(), OLEError> {
```

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,92 @@
...
+    F: Field + Serialize + Deserialize<'static>,
+    for<'a> OLEMessage<F>: Deserialize<'a>,
+{
+    async fn receive(&mut self, ctx: &mut C, b_k: Vec<F>) -> Result<Vec<F>, OLEError> {
+        let (receiver_adjust, adjust) = self.core.adjust(b_k).ok_or(OLEError::InsufficientOLEs)?;
+
+        let (_, adjust) = try_join!(
+            ctx,
+            ctx.io_mut().send(OLEMessage::Adjust(adjust)),
+            ctx.io_mut().expect_next::<OLEMessage<F>>()
```
sinui0: Not worth forking the thread for IO. You can just push the message into the IO buffer and then await the expected message which will potentially already be sitting in the IO buffer.

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,92 @@
+use crate::{msg::OLEMessage, OLEError, OLEReceiver as OLEReceive};
+use async_trait::async_trait;
+use itybity::ToBits;
+use mpz_common::{try_join, Context};
+use mpz_fields::Field;
+use mpz_ole_core::OLEReceiver as OLECoreReceiver;
+use mpz_ot::RandomOTReceiver;
+use serde::{Deserialize, Serialize};
```
sinui0: You can use `serio::{Serialize, Deserialize}` bounds

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,92 @@
...
+use serio::stream::IoStreamExt;
+use serio::SinkExt;
+use std::marker::PhantomData;
+
+/// OLE receiver.
+pub struct OLEReceiver<const M: usize, const N: usize, T, F, C> {
```
sinui0: I feel we can get rid of the const noise in the signature using associated types.

Eg.

```rust
// naming could use some thought
pub trait ByteRepr {
  type Serialized: AsRef<[u8]>;
}

impl ByteRepr for P256 {
  type Serialized = [u8; 32];
}
```

then use that bound elsewhere

```rust
impl<F: Field + ByteRepr>
```

Or just add it as a bound on `Field`

sinui0 (changes requested):

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+pub mod msg;
+pub mod rot;
+
+/// Batch OLE Sender.
+///
+/// The sender inputs field elements `a_k` and gets outputs `x_k`, such that
+/// `y_k = a_k * b_k + x_k` holds, where `b_k` and `y_k` are the [`OLEReceiver`]'s inputs and outputs
+/// respectively.
+#[async_trait]
+pub trait OLESender<C: Context, F: Field> {
```
sinui0: Let's be consistent with abbreviation across crates

```suggestion
pub trait OLESender<Ctx: Context, F: Field> {
```

crates/mpz-ole/src/rot/receiver.rs
th4s: Omg this works. I did not see it . The problem is that the `rot_receiver` depends on `Context`. Where clauses are so powerful.
```rust
impl<const M: usize, const N: usize, T, F> OLEReceiver<M, N, T, F>
where
    F: Field + Serialize + Deserialize,
{
    /// Preprocesses OLEs.
    ///
    /// # Arguments
    ///
    /// * `count` - The number of OLEs to preprocess.
    pub async fn preprocess<Ctx: Context>(
        &mut self,
        ctx: &mut Ctx,
        count: usize,
    ) -> Result<(), OLEError>
    where
        T: RandomOTReceiver<Ctx, bool, [u8; M]> + Send,
    {
        //...
    }
```

crates/mpz-ole/src/lib.rs
th4s: Interesting. I will try it out. What advantages do you see with this approach?

crates/mpz-ole/src/ideal.rs
sinui0, line 1: Let's stick to using the abstraction provided in `mpz-common` for ideal 2PC. Otherwise we end up with a lot of maintenance messing around with channels

crates/mpz-ole/src/ideal.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let ole_sender = IdealOLESender { channel: sender };
+
+    let ole_receiver = IdealOLEReceiver { channel: receiver };
+
+    (ole_sender, ole_receiver)
+}
+
+/// An ideal OLE Sender.
+pub struct IdealOLESender<F: Field> {
```
sinui0: Trait bounds on structs should be avoided if possible

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+/// `y_k = a_k * b_k + x_k` holds, where `b_k` and `y_k` are the [`OLEReceiver`]'s inputs and outputs
+/// respectively.
+#[async_trait]
+pub trait OLESender<Ctx: Context, F: Field> {
+    /// Sends his masked inputs to the [`OLEReceiver`].
+    ///
+    /// # Arguments
+    ///
+    /// * `ctx` - The context, which provides IO channels.
+    /// * `a_k` - The sender's OLE inputs.
```
sinui0: `a_k` is a messy argument name, I think we should stick to `inputs` and the user can refer to the documented functionality to understand the relation

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+/// The sender inputs field elements `a_k` and gets outputs `x_k`, such that
+/// `y_k = a_k * b_k + x_k` holds, where `b_k` and `y_k` are the [`OLEReceiver`]'s inputs and outputs
+/// respectively.
+#[async_trait]
+pub trait OLESender<Ctx: Context, F: Field> {
+    /// Sends his masked inputs to the [`OLEReceiver`].
+    ///
+    /// # Arguments
+    ///
+    /// * `ctx` - The context, which provides IO channels.
```
sinui0: Unnecessary detail

```suggestion
    /// * `ctx` - The context.
```

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+impl Display for OLEErrorKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            OLEErrorKind::OT => write!(f, "OT Error"),
+            OLEErrorKind::IO => write!(f, "IO Error"),
+            OLEErrorKind::Core => write!(f, "OLE Core Error"),
+            OLEErrorKind::InsufficientOLEs => write!(f, "Insufficient OLEs"),
+        }
+    }
+}
```
sinui0: I think we can do something like this:

```suggestion
impl Display for OLEError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self.kind {
            OLEErrorKind::OT => f.write_str("ot error"),
            ...
        }?;
        
        if let Some(source) = self.source.as_ref() {
          f.write_str(format!(" caused by: {source}"))?;
        }
        
        Ok(())
    }
}
```

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    source: Option<Box<dyn Error + Send + Sync>>,
+}
+
+impl OLEError {
+    fn new(kind: OLEErrorKind, source: Box<dyn Error + Send + Sync + 'static>) -> Self {
+        Self {
+            kind,
+            source: Some(source),
+        }
+    }
```
sinui0: For convenience:

```suggestion
    fn new<E>(kind: OLEErrorKind, source: E) -> Self 
    where
      E: Into<Box<dyn Error + Send + Sync>>,
    {
        Self {
            kind,
            source: Some(source.into()),
        }
    }
```

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,97 @@
...
+impl<T, F, Ctx: Context> OLEReceive<Ctx, F> for OLEReceiver<T, F>
+where
+    T: RandomOTReceiver<Ctx, bool, F::Serialized> + Send,
+    F: Field + Serialize + Deserialize,
+{
+    async fn receive(&mut self, ctx: &mut Ctx, b_k: Vec<F>) -> Result<Vec<F>, OLEError> {
+        let (receiver_adjust, adjust) = self.core.adjust(b_k).ok_or_else(|| {
+            OLEError::new(
+                OLEErrorKind::InsufficientOLEs,
+                "Not enough OLEs available".into(),
```
sinui0: lets include the number of OLEs there are, something like: `format!("{} < {}", self.core.len(), b_k.len())`

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        self.core.preprocess(rot_choices, rot_msg, masks)?;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl<T, F, Ctx: Context> OLEReceive<Ctx, F> for OLEReceiver<T, F>
+where
+    T: RandomOTReceiver<Ctx, bool, F::Serialized> + Send,
```
sinui0: Bound isn't needed here

crates/mpz-ole/src/rot/receiver.rs
```diff
@@ -0,0 +1,97 @@
...
+        let random_ot = self
+            .rot_receiver
+            .receive_random(ctx, count * F::BIT_SIZE as usize)
+            .await?;
+
+        let rot_msg: Vec<F> = random_ot
+            .msgs
+            .iter()
+            .map(|f| F::from_lsb0_iter(f.as_ref().iter_lsb0()))
+            .collect();
```
sinui0: Converting via the bit iterators is clunky. Let's either do it via an array conversion method, or change the OT bound to `RandomOTReceiver<Ctx, bool, F>` and find a way to have that available

crates/mpz-ole/src/rot/sender.rs
```diff
@@ -0,0 +1,101 @@
...
+        let random_ot = self
+            .rot_sender
+            .send_random(ctx, count * F::BIT_SIZE as usize)
+            .await?
+            .msgs
+            .iter()
+            .map(|[a, b]| {
+                [
+                    F::from_lsb0_iter(a.as_ref().into_iter_lsb0()),
+                    F::from_lsb0_iter(b.as_ref().into_iter_lsb0()),
```
sinui0: same here

sinui0 (changes requested):
So clean :ok_hand: 

Couple requests

crates/mpz-ole/src/ideal.rs
sinui0, line 1: Please add newtypes here, eg:

```rust
pub struct IdealOleSender(Alice<()>);
pub struct IdealOleReceiver(Bob<()>);
```

crates/mpz-ole/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    Field,
+    InsufficientOLEs,
+}
+
+impl From<OTError> for OLEError {
+    fn from(value: OTError) -> Self {
+        Self {
+            kind: OLEErrorKind::OT,
+            source: Some(Box::new(value) as Box<dyn Error + Send + Sync>),
+        }
```
sinui0: Doesn't matter, but you could use `Self::new(OLEErrorKind::OT, value)`

sinui0 (approved):
:fire: :fire: 

Just 1 nit on needing newtypes for the ideal impl.

themighty1 (approved):
gw!

PR #139, opened 2024-05-22
Introduce `hybrid-array` for OLE

th4s:
This PR gets rid of const generics in our OLE crates.

- It introduces an associated type number (typenum) for our fields.
- Use this typenum together with `hybrid-array` in `mpz-ole-core` and `mpz-ole`.


sinui0:
Would it be possible to include these changes into your other PRs instead of dedicated? I do think `hybrid-array` is a good approach

th4s:
Sure

th4s:
Added as commits into other PRs.

PR #140, opened 2024-05-23
refactor(mpz-garble-core): batched garbling

sinui0:
This PR refactors `mpz-garble-core` and implements stack-allocated batching. On my machine these changes result in ~50% improved garbling performance.

# Changes

- Reduce allocation overhead by reusing the memory between garblings.
- Replace `Option<Label>` with just `Label` and use default. This removes a bunch of branching to check whether the value is set.
- Add stack allocated batching of the encrypted gates
- Tweaked the control flow for iterating over the gates


crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+
+        let outputs = self
+            .outputs
             .iter()
             .map(|output| {
-                let labels: Vec<Label> = output
-                    .iter()
-                    .map(|node| self.active_labels[node.id()].expect("feed should be initialized"))
-                    .collect();
+                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```
themighty1, line 266: here `node` is Node<Feed> whereas it should be Node<Sink>. Maybe we could add an `iter_sink` to `BInaryRepr` or some other solution?

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+where
+    I: Iterator<Item = &'a Gate>,
+{
+    /// Enables hashing of the encrypted gates.
+    pub fn enable_hasher(&mut self) {
+        self.0.enable_hasher()
+    }
+
+    /// Returns `true` if the evaluator wants more encrypted gates.
+    pub fn wants_gates(&self) -> bool {
```
themighty1, line 299: this is never used

crates/mpz-garble-core/src/lib.rs
```diff
@@ -59,21 +53,40 @@ mod evaluator;
...
+    EncryptedGateBatchConsumer, EncryptedGateConsumer, Evaluator, EvaluatorError, EvaluatorOutput,
+};
+pub use generator::{
+    EncryptedGateBatchIter, EncryptedGateIter, Generator, GeneratorError, GeneratorOutput,
+};
+
+const KB: usize = 1024;
+const BYTES_PER_GATE: usize = 32;
+
+/// Maximum size of a batch.
```
themighty1:
```suggestion
/// Maximum size of a batch in bytes.
```

crates/mpz-garble-core/src/lib.rs
```diff
@@ -59,21 +53,40 @@ mod evaluator;
...
+const MAX_BATCH_SIZE: usize = 4 * KB;
+
+/// Default batch size.
```
themighty1:
```suggestion
/// Default amount of encrypted gates per batch.
```

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+        self.0.wants_gates()
+    }
+
+    /// Evaluates the next batch of gates in the circuit.
+    #[inline]
+    pub fn next(&mut self, batch: EncryptedGateBatch<N>) {
+        for encrypted_gate in batch.into_array() {
+            self.0.next(encrypted_gate);
+            if !self.0.wants_gates() {
+                return;
```
themighty1:
```suggestion
            if self.0.complete {
                // Skipping any remaining gates which may have been used to pad the last batch.
                return;
```

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+
+    /// Enables hashing of the encrypted gates.
+    pub fn enable_hasher(&mut self) {
+        self.hasher = Some(Hasher::new());
     }
 
-    /// Returns whether the generator has finished generating the circuit.
-    pub fn is_complete(&self) -> bool {
-        self.pos >= self.circ.gates().len()
+    /// Returns `true` if the generator has encrypted gates to send.
```
themighty1:
```suggestion
    /// Returns `true` if the generator has more encrypted gates to generate.
```

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+
+        let outputs = self
+            .outputs
             .iter()
             .map(|output| {
-                let labels: Vec<Label> = output
-                    .iter()
-                    .map(|node| self.low_labels[node.id()].expect("feed should be initialized"))
-                    .collect();
+                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```
themighty1, line 241: Node < Feed > should be Node < Sink >

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+where
+    I: Iterator<Item = &'a Gate>,
+{
+    /// Enables hashing of the encrypted gates.
+    pub fn enable_hasher(&mut self) {
+        self.0.enable_hasher()
+    }
+
+    /// Returns `true` if the generator has encrypted gates to send.
+    pub fn has_gates(&self) -> bool {
```
themighty1, line 335: should not be pub

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+impl<'a, I, const N: usize> EncryptedGateBatchIter<'a, I, N>
+where
+    I: Iterator<Item = &'a Gate>,
+{
+    /// Enables hashing of the encrypted gates.
+    pub fn enable_hasher(&mut self) {
+        self.0.enable_hasher()
+    }
+
+    /// Returns `true` if the generator has encrypted gates to send.
```
themighty1:
```suggestion
    /// Returns `true` if the generator has more encrypted gates to generate.
```

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+    /// Returns the encoded outputs of the circuit.
+    pub fn finish(mut self) -> Result<EvaluatorOutput, EvaluatorError> {
+        if self.wants_gates() {
             return Err(EvaluatorError::NotFinished);
         }
 
-        Ok(self
-            .circ
-            .outputs()
+        // Evaluate the remaining "free" gates.
```
themighty1: This is a special case which applies only if there are 0 AND gates, in all other cases if wants_gates() == false, it means that complete == true. Making this more explicit in the comment
```suggestion
        // If there were 0 AND gates in the circuit, we need to evaluate the "free" gates now.
```

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+    /// Returns the encoded outputs of the circuit, and the hash of the encrypted gates if present.
+    pub fn finish(mut self) -> Result<GeneratorOutput, GeneratorError> {
+        if self.has_gates() {
             return Err(GeneratorError::NotFinished);
         }
 
-        Ok(self
-            .circ
-            .outputs()
+        // Finish computing any "free" gates.
```
themighty1, line 232: Like above, this is a special case for 0 AND gates.
```suggestion
        // If there were 0 AND gates in the circuit, we compute any "free" gates now.
```

themighty1 (changes requested):
gw! we need to address the ```Node <Feed> / <Sink>``` matter

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+    cipher: &'static FixedKeyAes,
+    /// Buffer for the active labels.
+    labels: &'a mut [Label],
+    /// Iterator over the gates.
+    gates: I,
+    /// Circuit outputs.
+    outputs: &'a [BinaryRepr],
+    /// Current gate id.
+    gid: usize,
+    /// Hasher to use to hash the encrypted gates
```
th4s:
```suggestion
    /// Hasher to use to hash the encrypted gates.
```

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+        circ: &'a Circuit,
+        inputs: Vec<EncodedValue<state::Active>>,
+    ) -> Result<EncryptedGateBatchConsumer<'_, std::slice::Iter<'_, Gate>>, EvaluatorError> {
+        self.evaluate(circ, inputs).map(EncryptedGateBatchConsumer)
+    }
+}
+
+/// Consumer over the encrypted gates of a circuit.
+pub struct EncryptedGateConsumer<'a, I: Iterator> {
+    /// Cipher to use to encrypt the gates
```
th4s:
```suggestion
    /// Cipher to use to encrypt the gates.
```

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
-        while self.pos < self.circ.gates().len() {
-            match &self.circ.gates()[self.pos] {
-                Gate::Inv {
-                    x: node_x,
-                    z: node_z,
-                } => {
-                    let x = labels[node_x.id()].expect("feed should be initialized");
-                    labels[node_z.id()] = Some(x);
-                }
+    /// Evaluates the next gate in the circuit.
```
th4s: This comment could be better and explain that this function evaluates several gates up to the next AND-gate including it. Currently it states that it evaluates the next gate in the circuit, which is not quite correct.

crates/mpz-garble-core/src/evaluator.rs
th4s: I think `self.wants_gates() == false` only means that the circuit does not contain any more AND gates. But it is still possible that the tail contains non-evaluated XOR, INV gates, which need one more call to `next` so that `self.complete == true`.

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+        inputs: Vec<EncodedValue<state::Full>>,
+    ) -> Result<EncryptedGateBatchIter<'_, std::slice::Iter<'_, Gate>>, GeneratorError> {
+        self.generate(circ, delta, inputs)
+            .map(EncryptedGateBatchIter)
+    }
+}
+
+/// Iterator over encrypted gates of a garbled circuit.
+pub struct EncryptedGateIter<'a, I> {
+    /// Cipher to use to encrypt the gates
```
th4s:
```suggestion
    /// Cipher to use to encrypt the gates.
```

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+    delta: Delta,
+    /// Buffer for the 0-bit labels.
+    labels: &'a mut [Label],
+    /// Iterator over the gates.
+    gates: I,
+    /// Circuit outputs.
+    outputs: &'a [BinaryRepr],
+    /// Current gate id.
+    gid: usize,
+    /// Hasher to use to hash the encrypted gates
```
th4s:
```suggestion
    /// Hasher to use to hash the encrypted gates.
```

th4s (commented):
lgtm :+1:

crates/mpz-garble-core/src/evaluator.rs
sinui0: This is a detail of `mpz-circuits` and out of scope, I've not changed this here. The `Node` newtype is mostly for the circuit builder API.

crates/mpz-garble-core/src/evaluator.rs
sinui0: It is used when IO is added

crates/mpz-garble-core/src/evaluator.rs
sinui0: `self.0.complete` is not the correct thing to check here, as there may be additional XOR/INV gates to process. This just makes sure that we don't pass in any padding gates

crates/mpz-garble-core/src/generator.rs
sinui0: Same comment as above, this is out of scope

crates/mpz-garble-core/src/generator.rs
sinui0: It's `pub` on purpose, this can be useful in the IO wrapper

crates/mpz-garble-core/src/evaluator.rs
sinui0: That's an internal implementation detail the user doesn't need to be aware of. I will qualify it as the `next encrypted gate`

crates/mpz-garble-core/src/generator.rs
sinui0: @th4s  answered this above

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+
+        let outputs = self
+            .outputs
             .iter()
             .map(|output| {
-                let labels: Vec<Label> = output
-                    .iter()
-                    .map(|node| self.active_labels[node.id()].expect("feed should be initialized"))
-                    .collect();
+                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```
themighty1, line 266:
```suggestion
                // XXX: make a change in `mpz-circuits` to be able to iterate over sink nodes here.
                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```

crates/mpz-garble-core/src/evaluator.rs
themighty1: ok, in my new review i'll suggest a comment to add here to show that it is not an error

crates/mpz-garble-core/src/generator.rs
```diff
@@ -140,111 +126,254 @@ impl Generator {
...
+
+        let outputs = self
+            .outputs
             .iter()
             .map(|output| {
-                let labels: Vec<Label> = output
-                    .iter()
-                    .map(|node| self.low_labels[node.id()].expect("feed should be initialized"))
-                    .collect();
+                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```
themighty1, line 241:
```suggestion
                // XXX: make a change in `mpz-circuits` to be able to iterate over sink nodes here.
                let labels: Vec<Label> = output.iter().map(|node| self.labels[node.id()]).collect();
```

crates/mpz-garble-core/src/evaluator.rs
themighty1: ok, i'll just copy the comment line into the new review

crates/mpz-garble-core/src/evaluator.rs
```diff
@@ -122,111 +115,204 @@ impl Evaluator {
...
+        self.0.wants_gates()
+    }
+
+    /// Evaluates the next batch of gates in the circuit.
+    #[inline]
+    pub fn next(&mut self, batch: EncryptedGateBatch<N>) {
+        for encrypted_gate in batch.into_array() {
+            self.0.next(encrypted_gate);
+            if !self.0.wants_gates() {
+                return;
```
themighty1, line 310:
```suggestion
                // Skipping any remaining gates which may have been used to pad the last batch.
                return;
```

crates/mpz-garble-core/src/evaluator.rs
themighty1: My suggested comment is still valid, ill explain on the call today.

crates/mpz-garble-core/src/generator.rs
themighty1: My suggested comment is still valid, ill explain on the call today.

themighty1 (approved):
I added a comment about sink nodes to make it clear to the reader that there is no error there.

crates/mpz-garble-core/src/evaluator.rs
themighty1: That's not how I see the code works, I'll explain on the call today.

PR #141, opened 2024-05-24
feat(mpz-common): Context::blocking

sinui0:
This PR adds the `Context::blocking` method which supports temporarily moving a context to a separate thread to perform blocking computation. I've also moved the CPU backend shim abstraction into `mpz-common`.

This method is needed to avoid deadlocks on single threaded machines. The alternative approach is to use an `mpsc` channel between the worker thread and the calling context, but that doesn't work if single-threaded.

### DIFF
--- a/crates/mpz-common/src/cpu.rs
+++ b/crates/mpz-common/src/cpu.rs
@@ -1,0 +1,122 @@
+//! CPU backend shim.
+
+use cfg_if::cfg_if;
+
+cfg_if! {
+    if #[cfg(feature = "force-st")] {
+        pub use st::SingleThreadedBackend as CpuBackend;
+    } else if #[cfg(feature = "rayon")] {
+        pub use rayon_backend::RayonBackend as CpuBackend;
+    } else {
+        pub use st::SingleThreadedBackend as CpuBackend;
+    }
+}
+
+#[cfg(any(feature = "force-st", not(feature = "rayon")))]
+mod st {
+    use futures::Future;
+
+    /// A single-threaded CPU backend.
+    #[derive(Debug)]
+    pub struct SingleThreadedBackend;
+
+    impl SingleThreadedBackend {
+        /// Execute a future on the CPU backend.
+        #[inline]
+        pub fn blocking_async<F>(fut: F) -> impl Future<Output = F::Output> + Send
+        where
+            F: Future + Send + 'static,
+            F::Output: Send,
+        {
+            fut
+        }
+
+        /// Execute a closure on the CPU backend.
+        #[inline]
+        pub fn blocking<F, R>(f: F) -> impl Future<Output = R> + Send
+        where
+            F: FnOnce() -> R + Send + 'static,
+            R: Send + 'static,
+        {
+            async move { f() }
+        }
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+        use pollster::block_on;
+
+        #[test]
+        fn test_st_backend_blocking() {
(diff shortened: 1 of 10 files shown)


th4s (commented):
Hmm, I am not sure I understand this PR. A few questions:

> This method is needed to avoid deadlocks on single threaded machines.

What is a single-threaded machine and why do we need to support single-threaded machines? Or do you mean a machine with a single CPU core?

> The alternative approach is to use an mpsc channel between the worker thread and the calling context, but that doesn't work if single-threaded.

Why doesn't this work?

sinui0:
> What is a single-threaded machine and why do we need to support single-threaded machines? Or do you mean a machine with a single CPU core?

Yes, I'm referring to single-core CPUs, and generally environments without threading. Not to be confused with our logical thread abstraction in `mpz-common`.

> Why doesn't this work?

It does work for `CpuBackend::blocking_async` but not for `CpuBackend::blocking`. Hard to point out exactly why, but it leads to a deadlock.

---

The primary purpose of the `Context::blocking` method is that it handles moving a context to another thread in multi-threaded case. This isn't otherwise possible because most callers are going to be working with a `&mut impl Context` which obviously isn't `'static`.

crates/mpz-common/src/cpu.rs
```diff
@@ -0,0 +1,122 @@
...
+#[cfg(any(feature = "force-st", not(feature = "rayon")))]
+mod st {
+    use futures::Future;
+
+    /// A single-threaded CPU backend.
+    #[derive(Debug)]
+    pub struct SingleThreadedBackend;
+
+    impl SingleThreadedBackend {
+        /// Execute a future on the CPU backend.
```
themighty1:
```suggestion
        /// Executes a future on the CPU backend.
```

crates/mpz-common/src/cpu.rs
```diff
@@ -0,0 +1,122 @@
...
+        #[inline]
+        pub fn blocking_async<F>(fut: F) -> impl Future<Output = F::Output> + Send
+        where
+            F: Future + Send + 'static,
+            F::Output: Send,
+        {
+            fut
+        }
+
+        /// Execute a closure on the CPU backend.
```
themighty1:
```suggestion
        /// Executes a closure on the CPU backend.
```

crates/mpz-common/src/cpu.rs
```diff
@@ -0,0 +1,122 @@
...
+mod rayon_backend {
+    use futures::{channel::oneshot, Future};
+    use pollster::block_on;
+
+    /// A Rayon CPU backend.
+    #[derive(Debug)]
+    pub struct RayonBackend;
+
+    impl RayonBackend {
+        /// Execute a future on the CPU backend.
```
themighty1:
```suggestion
        /// Executes a future on the CPU backend.
```

crates/mpz-common/src/cpu.rs
```diff
@@ -0,0 +1,122 @@
...
+                let (sender, receiver) = oneshot::channel();
+                rayon::spawn(move || {
+                    let output = block_on(fut);
+                    _ = sender.send(output);
+                });
+                receiver.await.expect("worker thread does not drop channel")
+            }
+        }
+
+        /// Execute a closure on the CPU backend.
```
themighty1:
```suggestion
        /// Executes a closure on the CPU backend.
```

themighty1 (approved):

th4s (approved):

PR #143, opened 2024-05-29
feat(mpz-common): scoped! macro

sinui0:
This PR adds a `scoped!` macro which makes it easy to write closures which return scoped futures. This macro may end up being preferable to using the `join` and `try_join` macros.


sinui0:
I look forward to the resolution of https://github.com/rust-lang/rust/issues/70263 which will let us get rid of this and associated utilities.

th4s (approved):
:+1:

themighty1 (approved):

Issue #112, opened 2024-03-07
Tracking: Threading + IO redesign


sinui0:
This is a tracking issue for the redesign work for how we model protocol execution and IO.

[Branch](https://github.com/privacy-scaling-explorations/mpz/tree/threading-refactor)

## Tasks

- [x] Introduce `mpz-common` which provides types and traits for threading and synchronization. It also adds [`serio`](https://github.com/tlsnotary/tlsn-utils/tree/dev/serio) for our IO abstraction. #107 
- [x] Encapsulate our coin-toss protocols async layer #108 
- [x] Update `mpz-ot` to use `mpz-common` and deprecate the KOS15 actor impl #109 
- [x] Finish implementing OT traits for KOS15 shared variant.
- [x] Implement a multi-threaded executor/context in `mpz-common`
- [x] Update `mpz-garble`
- [x] Update `mpz-share-conversion`

PR #144, opened 2024-05-29
refactor(mpz-garble): fix threading breaking changes

sinui0:
This PR refactors `mpz-garble` to accomodate the breaking changes from #112 .

# Changes
- Use new `Context` abstraction.
- Incorporate changes from #140 
- Remove `DEAPVm` type and just use `DEAPThread`

### DIFF
--- a/crates/mpz-garble/src/generator/mod.rs
+++ b/crates/mpz-garble/src/generator/mod.rs
@@ -9,17 +9,18 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-use futures::{Sink, SinkExt};
 use mpz_circuits::{
     types::{Value, ValueType},
     Circuit,
 };
+use mpz_common::{scoped, Context};
 use mpz_core::hash::Hash;
 use mpz_garble_core::{
-    encoding_state, msg::GarbleMessage, ChaChaEncoder, EncodedValue, Encoder,
-    Generator as GeneratorCore,
+    encoding_state, ChaChaEncoder, EncodedValue, Encoder, EncodingCommitment,
+    Generator as GeneratorCore, GeneratorOutput,
 };
-use utils_aio::non_blocking_backend::{Backend, NonBlockingBackend};
+use serio::SinkExt;
+use tracing::{span, Level};
 
 use crate::{
     memory::EncodingMemory,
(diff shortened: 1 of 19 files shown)


th4s (approved):
lgtm :+1:

crates/mpz-garble/src/generator/mod.rs
```diff
@@ -298,41 +303,40 @@ impl Generator {
...
+                let mut gen = GeneratorCore::default();
+                let mut gen_iter = gen.generate_batched(&circ, delta, inputs)?;
+                let io = ctx.io_mut();
+
+                if hash {
+                    gen_iter.enable_hasher();
+                }
+
+                while let Some(batch) = gen_iter.by_ref().next() {
+                    io.feed(batch).await?;
```
themighty1, line 323: Are we sure we need `feed` here instead of `send`? WIth `feed`, the gates will not be streamed until `flush` is called.

crates/mpz-garble/src/protocol/deap/vm.rs
```diff
@@ -1,347 +1,248 @@
...
-            _role: role,
-            op_id,
-            deap,
-            ot_send,
-            ot_recv,
-            sink,
-            stream,
+    /// Finalizes the DEAP instance.
+    ///
+    /// If this instance is the leader this function returns the follower's
```
themighty1:
```suggestion
    /// If this instance is the leader, this function returns the follower's
```

themighty1 (changes requested):
nice work!!! just one concern about feed vs send

crates/mpz-garble/src/generator/mod.rs
sinui0: @th4s and I discussed this and think that `feed` is correct here. We don't want to wait for the IO to flush before continuing to garble the next batch. There is a latency vs throughput tradeoff here, where worst case latency occurs when the garbled circuit is smaller than the IO buffer so it doesn't flush until garbling is complete. Otherwise, calling `feed` _will_ start flushing once the buffer is full. In our case the underlying connection is being polled concurrently so it will make progress between calls to `feed`.

PR #147, opened 2024-06-03
Add `mpz-share-conversion-core`

th4s:
This PR reimplements the `mpz-share-conversion-core` crate, which provides A2M and M2A. It also removes `mpz-share-conversion` for now.


crates/mpz-share-conversion-core/src/ideal.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Creates a new functionality.
+    pub fn new() -> Self {
+        Self(thread_rng())
+    }
+
+    /// Generates additive shares from multiplicative shares.
+    pub fn generate<F: Field>(
+        &mut self,
+        sender_input: &[F],
+        receiver_input: &[F],
```
sinui0:
```suggestion
        sender_input: Vec<F>,
        receiver_input: Vec<F>,
```

crates/mpz-share-conversion-core/src/ideal.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Creates a new functionality.
+    pub fn new() -> Self {
+        Self(thread_rng())
+    }
+
+    /// Generates multiplicative shares from additive shares.
+    pub fn generate<F: Field>(
+        &mut self,
+        sender_input: &[F],
+        receiver_input: &[F],
```
sinui0:
```suggestion
        sender_input: Vec<F>,
        receiver_input: Vec<F>,
```

crates/mpz-share-conversion-core/src/ideal.rs
```diff
@@ -0,0 +1,134 @@
+//! Ideal functionalities for share conversion.
+
+use mpz_fields::Field;
+use rand::{rngs::ThreadRng, thread_rng};
+
+/// The M2A functionality.
+pub struct IdealM2A(ThreadRng);
```
sinui0: `ThreadRng` is `!Send` so will cause sadness in async code

crates/mpz-share-conversion-core/src/lib.rs
```diff
@@ -1,77 +1,56 @@
...
-                    }
-                },
-            )
-            .collect()
+        Ok(())
     }
 }
+
+#[derive(Debug)]
+pub(crate) enum ShareConversionErrorKind {
```
sinui0:
```suggestion
pub(crate) enum ErrorKind {
```

sinui0 (approved):
LGTM, couple nits

crates/mpz-share-conversion-core/src/a2m.rs
```diff
@@ -0,0 +1,124 @@
...
+    ole_input.iter_mut().for_each(|r| *r = r.inverse());
+
+    Ok((ole_input, A2MMasks(masks)))
+}
+
+/// Converts the A2M sender's masks into multiplicative receiver shares.
+///
+/// # Arguments
+///
+/// * `masks` - The masks received by the sender.
```
themighty1:
```suggestion
/// * `masks` - The masks received from the sender.
```

crates/mpz-share-conversion-core/src/ideal.rs
```diff
@@ -0,0 +1,116 @@
...
+        sender_input: Vec<F>,
+        receiver_input: Vec<F>,
+    ) -> (Vec<F>, Vec<F>) {
+        assert_eq!(
+            sender_input.len(),
+            receiver_input.len(),
+            "Vectors of field elements should have equal length."
+        );
+
+        let mut rng = thread_rng();
```
themighty1: we try to achieve deterministic tests, so we use        let mut rng = Prg::from_seed(Block::ZERO);

crates/mpz-share-conversion-core/src/ideal.rs
```diff
@@ -0,0 +1,116 @@
...
```
themighty1: let mut rng = Prg::from_seed(Block::ZERO);

themighty1 (approved):
gw!!!

crates/mpz-share-conversion-core/src/ideal.rs
th4s: Hmm but these are the ideal implementations. Do we want this here? Or do we want this only for input in test cases?

sinui0: Ideal impls can probably just use fixed seeds like Dan suggests

th4s: What if we have a protocol we want to test, which depends on let's say 2 other ideal functionalities. When now using `Prg::from_seed(Block::ZERO)` in each of them, we have introduced a correlation, which is not there in the real implementation of said protocol. It is possible that these correlations would even break the protocol when testing it, for example because they cancel out when used in a subtraction.

Is this example too much made up?

themighty1: That's a fair point. Let's not use `Block::ZERO` everywhere but some other nothing-up-my-sleeve deterministic value like eg "05/06/2025" <- the date when the test was added for example.

themighty1: I meant to say, it is still OK to use 0,1,2,3,4... to seed PRGs. But for cases where doing so may potentially lead to side effect like in ideal functionalities, we may use more unique values like dates.

PR #146, opened 2024-05-31
refactor(mpz-share-conversion): new impl

sinui0:
This PR stubs out the traits for `mpz-share-conversion` to accomodate #112

PR #148, opened 2024-06-03
Add implementations for A2M/M2A

th4s:
This PR adds implementations of the traits A2M/M2A.


sinui0:
Merging into #146 and combining review there

PR #149, opened 2024-06-06
feat(mpz-garble): pre-commit inputs

sinui0:
This PR exposes the ability to commit assigned values in memory.

### DIFF
--- a/crates/mpz-garble/src/protocol/deap/mod.rs
+++ b/crates/mpz-garble/src/protocol/deap/mod.rs
@@ -153,12 +153,6 @@ impl DEAP {
         OTR: OTReceiveEncoding<Ctx> + Send,
     {
         let assigned = self.state().memory.drain_assigned(inputs);
-        if assigned.len() != inputs.iter().map(|value| value.len()).sum() {
-            return Err(DEAPError::CommitmentError(
-                "attempted to commit unassigned values".to_string(),
-            ));
-        }
-
         match self.role {
             Role::Leader => {
                 try_join!(
@@ -199,17 +193,6 @@ impl DEAP {
         OTR: OTReceiveEncoding<Ctx> + Send,
     {
         let assigned = self.state().memory.drain_assigned(values);
-
-        if assigned.len() != values.iter().map(|value| value.len()).sum() {
-            return Err(DEAPError::CommitmentError(
-                "attempted to commit unassigned values".to_string(),
-            ));
-        } else if !assigned.blind.is_empty() {
-            return Err(DEAPError::CommitmentError(
-                "attempted to commit blind values as prover".to_string(),
-            ));
-        }
-
         self.ev
             .setup_assigned_values(ctx, &assigned, ot_recv)
             .await?;
@@ -229,17 +212,6 @@ impl DEAP {
         OTS: OTSendEncoding<Ctx> + Send,
     {
         let assigned = self.state().memory.drain_assigned(values);
-
-        if assigned.len() != values.iter().map(|value| value.len()).sum() {
-            return Err(DEAPError::CommitmentError(
-                "attempted to commit to unassigned values".to_string(),
-            ));
-        } else if !assigned.private.is_empty() {
-            return Err(DEAPError::CommitmentError(
-                "attempted to commit private values as verifier".to_string(),
-            ));
-        }
-
         self.gen
             .setup_assigned_values(ctx, &assigned, ot_send)
             .await?;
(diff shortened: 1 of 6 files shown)


th4s (approved):

crates/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -139,6 +139,86 @@ impl DEAP {
...
+        inputs: &[ValueRef],
+        ot_send: &mut OTS,
+        ot_recv: &mut OTR,
+    ) -> Result<(), DEAPError>
+    where
+        Ctx: Context,
+        OTS: OTSendEncoding<Ctx> + Send,
+        OTR: OTReceiveEncoding<Ctx> + Send,
+    {
+        let assigned = self.state().memory.drain_assigned(inputs);
```
themighty1, line 157: Seems like a good idea to check here and return an error if the `inputs` were not assigned prior to calling `commit` 
and also for the 2 methods below

crates/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -139,6 +139,86 @@ impl DEAP {
...
+    {
+        let assigned = self.state().memory.drain_assigned(values);
+        self.ev
+            .setup_assigned_values(ctx, &assigned, ot_recv)
+            .await?;
+
+        Ok(())
+    }
+
+    /// Commits the provided values for verifying.
```
themighty1: Reading the comment may seem confusing since the verifier is not actually committing to anything, instead maybe 
```suggestion
    /// Enables the prover to commit the provided values for verifying.
```

crates/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -139,6 +139,86 @@ impl DEAP {
...
+        &self,
+        ctx: &mut Ctx,
+        values: &[ValueRef],
+        ot_recv: &mut OTR,
+    ) -> Result<(), DEAPError>
+    where
+        Ctx: Context,
+        OTR: OTReceiveEncoding<Ctx> + Send,
+    {
+        let assigned = self.state().memory.drain_assigned(values);
```
themighty1, line 199: Can we have a check that only private values can be committed to for proving?

crates/mpz-garble/src/protocol/deap/mod.rs
```diff
@@ -139,6 +139,86 @@ impl DEAP {
...
+        &self,
+        ctx: &mut Ctx,
+        values: &[ValueRef],
+        ot_send: &mut OTS,
+    ) -> Result<(), DEAPError>
+    where
+        Ctx: Context,
+        OTS: OTSendEncoding<Ctx> + Send,
+    {
+        let assigned = self.state().memory.drain_assigned(values);
```
themighty1, line 220: Can we have a check that only blind values can be committed to for verification?

themighty1 (changes requested):

crates/mpz-garble/src/protocol/deap/mod.rs
sinui0: Think of it more like a "database commit" instead of cryptographic commitment. The two parties have a shared virtual machine and are jointly committing values to its memory.

themighty1: Ideally, we should avoid leaking the VM impl details into the DEAP protocol API, if possible.

The person who follows the DEAP spec doc will be confused by the sudden overload of the term "commit", imo. To avoid the confusion, we need to explain the term "commit" from the cryptographic commitment pov.

crates/mpz-garble/src/protocol/deap/mod.rs
themighty1: Based on our DMs, let's use the "commit if possible" approach. Then the check is not required. Let's add a clarifying comment here and in the 2 methods below like e.g.
```
/// 
/// If some of the values in `inputs` were committed earlier, this method will ignore them.
```

crates/mpz-garble/src/protocol/deap/mod.rs
themighty1: Based on our DMs adding these checks requires other changes also. This part of the code will be obsoleted soon, so we decided to not make any changes at this time.

crates/mpz-garble/src/protocol/deap/mod.rs
themighty1: Based on our DMs adding these checks requires other changes also. This part of the code will be obsoleted soon, so we decided to not make any changes at this time.

themighty1 (approved):
lgtm!
Feel free to address my comment suggestion as you deem best.

PR #155, opened 2024-06-10
refactor: KOS and preprocessing traits

sinui0:
This PR adds traits for the preprocessing model and also updates KOS to implement random OT for any `Standard: Distribution<T>` type.

`Allocate` trait rationale
---

It is difficult to properly model preprocessing in the case where a functionality is shared by multiple others. One approach would be to update the `preprocess` function so that it has a `count` argument then use some sort of async `Barrier` for synchronization. That approach doesn't work in the case where multiplexing isn't available (only 1 context).

The approach I settled on was to decouple reserving capacity step from execution of the preprocessing. The implemented model is 2 steps:

1. A functionality is injected as a sub-protocol into other protocols. Those protocols themselves implement `Allocate`, which allocate what they need from the sub-protocol. This process does not requires synchronization.
2. Once all protocols have allocated what they need, `Preprocess` is called which actually executes the preprocessing all batched together.

Modeling it this way works for both cases where a functionality is "shared" or owned directly by another protocol.

Side note: Once we have SoftSpoken which can be extended multiple times, we should be able to just share the base OT and perform OT extensions in parallel instead of this batching. This will remove the need for shared state synchronization and be more performant in terms of latency and compute.

### DIFF
--- a/crates/mpz-common/src/lib.rs
+++ b/crates/mpz-common/src/lib.rs
@@ -23,11 +23,28 @@ pub mod ideal;
 #[cfg(feature = "sync")]
 pub mod sync;
 
+use async_trait::async_trait;
 pub use context::{Context, ContextError};
 pub use id::{Counter, ThreadId};
 
 // Re-export scoped-futures for use with the callback-like API in `Context`.
 pub use scoped_futures;
+
+/// Allocates capacity from a functionality in the pre-processing model.
+pub trait Allocate {
+    /// Allocates `count` capacity.
+    fn alloc(&mut self, count: usize);
+}
+
+/// A functionality in the pre-processing model.
+#[async_trait]
+pub trait Preprocess<Ctx>: Allocate {
+    /// Error type.
+    type Error;
+
+    /// Preprocesses the functionality.
+    async fn preprocess(&mut self, ctx: &mut Ctx) -> Result<(), Self::Error>;
+}
 
 /// A convenience macro for creating a closure which returns a scoped future.
 ///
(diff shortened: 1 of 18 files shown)


crates/mpz-common/src/lib.rs
```diff
@@ -23,12 +23,29 @@ pub mod ideal;
...
+pub trait Allocate {
+    /// Allocates `count` capacity.
+    fn alloc(&mut self, count: usize);
+}
+
+/// A functionality in the pre-processing model.
+#[async_trait]
+pub trait Preprocess<Ctx>: Allocate {
+    /// Error type.
+    type Error;
```
th4s: Can we add a `std::error::Error` trait bound here? That usually makes it easier to work with.

crates/mpz-share-conversion/src/sender.rs
```diff
@@ -1,29 +1,66 @@
...
 use rand::thread_rng;
 use serio::{Deserialize, Serialize, SinkExt};
 use std::marker::PhantomData;
 
 /// Sender for share conversion.
+#[derive(Debug)]
 pub struct ShareConversionSender<T, F> {
     ole_sender: T,
-    phantom: PhantomData<F>,
+    _pd: PhantomData<F>,
```
th4s, line 15: Let's have a consistent name for `PhantomData` for sender and receiver.

th4s (approved):
Very nice :rocket:

themighty1 (approved):
ACK

PR #156, opened 2024-06-10
feat(mpz-circuits-generic): implement generic circuit struct

brech1:
# Description

This PR aims to introduce a new generic circuit struct:

```rust
pub struct Circuit<T> {
    input_count: usize,
    output_count: usize,
    gates: Vec<T>,
}
```

The `Circuit` is a collection of gates. Since for the execution of these gates is important that these form a directed acyclic graph (DAG), the correct way of building this circuit is using the `CircuitBuilder`

```rust
pub struct CircuitBuilder<T> {
    current_node: Node,
    inputs: Vec<Node>,
    outputs: Vec<Node>,
    gates: Vec<T>,
    stack_size: usize,
}
```

The `CircuitBuilder` API ensures that the gates are ordered topologically, this means in the order they should be executed taking dependencies into account

Here is our only constrain for the gates, they have to implement the `Component` trait, this means that they should be able to return an iterator over the nodes.

```rust
pub trait Component {
    fn get_inputs(&self) -> impl Iterator<Item = &Node>;
    fn get_outputs(&self) -> impl Iterator<Item = &Node>;
}
```

The `Node` struct:

```rust
pub struct Node(pub(crate) u32);
```

## Example

```rust
let mut builder = CircuitBuilder::<Gate>::new();

let (in_0, in_1) = (builder.add_input(), builder.add_input());

let &Gate { output, .. } = builder
    .add_gate(|next| Gate {
        inputs: vec![in_0, in_1],
        output: next.next(),
    })
    .unwrap();
    
let &Gate { output, .. } = builder
    .add_gate(|next| Gate {
        inputs: vec![in_0, output],
        output: next.next(),
    })
    .unwrap();

builder.add_output(output);
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "crates/mpz-common",
     "crates/mpz-fields",
     "crates/mpz-circuits",
+    "crates/mpz-circuits-generic",
     "crates/mpz-circuits-macros",
     "crates/mpz-cointoss",
     "crates/mpz-cointoss-core",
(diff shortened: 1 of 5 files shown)


sinui0:
is this awaiting review?

brech1:
> is this awaiting review?

Yes, I added the checks we discussed on our last call

brech1:
Closing this PR due to change in direction 🚀

PR #157, opened 2024-06-12
fix(mpz-common): flush io in syncer

sinui0:
This PR fixes a bug in the syncers. The leader *must* flush the io before returning, otherwise it deadlocks. Also, the leader should be sending the ticket *before* executing the function/future.

### DIFF
--- a/crates/mpz-common/src/sync/async_syncer.rs
+++ b/crates/mpz-common/src/sync/async_syncer.rs
@@ -5,8 +5,8 @@ use std::{
     task::{ready, Context as StdContext, Poll, Waker},
 };
 
-use futures::{future::poll_fn, Future};
-use serio::{stream::IoStreamExt, IoDuplex};
+use futures::Future;
+use serio::{stream::IoStreamExt, IoDuplex, SinkExt};
 use tokio::sync::Mutex;
 
 use crate::sync::{SyncError, Ticket};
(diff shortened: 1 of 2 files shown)


th4s (approved):

crates/mpz-common/src/sync/async_syncer.rs
```diff
@@ -67,15 +67,11 @@ impl Leader {
...
-        let mut io = Pin::new(io);
-        poll_fn(|cx| io.as_mut().poll_ready(cx)).await?;
-        let (output, tick) = {
-            let mut tick_lock = self.tick.lock().await;
-            let output = fut.await;
-            let tick = tick_lock.increment_in_place();
-            (output, tick)
-        };
-        io.start_send(tick)?;
+        let mut tick_lock = self.tick.lock().await;
```
themighty1, line 70: I don't quite understand why we don't release the lock on the ticket as soon as possible?
The way it is coded now will make other tasks wait until this task completes.
I expect this:
```rust
let mut tick_lock = self.tick.lock().await;
let tick = tick_lock.increment_in_place(); 
drop(tick_lock);
io.send(tick).await?;
let output = fut.await;
```

themighty1: Additionally, I think we want to `join!(io.send(tick), fut)` in order to avoid the overhead of awaiting `io.send`

themighty1 (changes requested):
just need to clarify a few points

crates/mpz-common/src/sync/async_syncer.rs
sinui0: The lock must be held until the task is finished otherwise competing CPU threads can budge in line. The purpose of the syncer is to make sure that any state mutation performed by a task is executed in the same order across machines.

I do agree, we can join on the futures for sending the ticket and executing the task. I'll update it

themighty1: Right, makes sense re the order of execution, thanks. Ok, lets wait for the update.

PR #159, opened 2024-06-14
Vope core and protocol with io

xiangxiecrypto:
- [x] VOPE Core protocol, consists of all the building blocks.
- [x] VOPE with io.

I created two `mpz-zk-core` and `mpz-zk` crates, and put VOPE to that crates. VOPE is a building block for QuickSilver.


xiangxiecrypto:
I will resolve the conflicts after ferret is merged.

PR #173, opened 2024-07-05
refactor(mpz-ot): ferret clean up

sinui0:
This PR does some simplification and clean up of the ferret impl.

# Changes

- Refactored to opaque error types
- Removed state from `spcot` and `mpcot` impls, making them functional
- Removed unnecessary generics
- Removed some unsatisfiable trait bounds, eg `RandomCOT: Default`
- Added buffering to the core types
- Implemented iterative extension with one-time bootstrapping using injected RCOT
  - This still needs work, from my understanding the iterations can be performed in parallel. Right now we're doing them sequentially and that will have poor round complexity.


crates/mpz-ot/src/ferret/mod.rs
```diff
@@ -63,21 +42,23 @@ impl<RandomCOT: Clone, SetupRandomCOT> FerretConfig<RandomCOT, SetupRandomCOT> {
     pub fn lpn_parameters(&self) -> LpnParameters {
         self.lpn_parameters
     }
+
+    /// Returns the number of COTs required to bootstrap an extension.
+    pub(crate) fn bootstrap_rate(&self) -> usize {
+        self.lpn_parameters.k as usize
+    }
```
xiangxiecrypto:
```suggestion
    /// Returns the number of COTs required to bootstrap an extension.
    pub(crate) fn bootstrap_rate(&self) -> usize {
        self.lpn_parameters.k
    }
```

crates/mpz-ot/src/ferret/mpcot.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[tokio::test]
+    async fn test_mpcot(#[case] lpn_type: LpnType) {
+        use crate::Correlation;
+
+        let (mut ctx_sender, mut ctx_receiver) = test_st_executor(8);
+        let (mut rcot_sender, mut rcot_receiver) = ideal_rcot();
+
+        let alphas = vec![0, 1, 3, 4, 2];
+        let t = alphas.len();
+        let n = 10;
```
xiangxiecrypto: The `alphas` values are not regular, so the test fails.
Try to use a regular `alphas` such as `[0, 3, 4, 7, 9]`

crates/mpz-ot/src/ferret/sender.rs
```diff
@@ -89,72 +82,122 @@ where
...
+                mpcot::send(
+                    ctx,
+                    &mut BootstrappedSender(&mut sender),
+                    delta,
+                    lpn_type,
+                    t,
+                    n,
+                )
+                .await?
+            };
```
xiangxiecrypto: I do not think we should use `self.rcot` to generate COTs in this step. It will only be used in the setup phase.
One optimization is to reserve enough COTs in the previous ferret extension for the next extension, and just read them for this step like `BootstrappedSender`.

xiangxiecrypto: Note that the `k` COTs are already reserved in Sender state (see `v`).

crates/mpz-ot/src/ferret/receiver.rs
```diff
@@ -96,97 +91,98 @@ where
...
+            } else {
+                mpcot::receive(
+                    ctx,
+                    &mut BootstrappedReceiver(&mut receiver),
+                    lpn_type,
+                    alphas,
+                    n as u32,
+                )
+                .await?
+            };
```
xiangxiecrypto: similar to Sender.

xiangxiecrypto (changes requested):
It is much simpler! LGTM

xiangxiecrypto:
Optimize the ferret extension process, rcot is only invoked in the setup phase, and buffer all the needed COTs in the extension phase.
@sinui0 Is there any example of using KOS with io? Then I can set all the real parameters for Ferret.

sinui0:
ty @xiangxiecrypto sorry this is taking so long on my end, im almost ready to focus back here again.

Why move the buffering out into the io layer? Can't we use the existing buffer in core?

> Is there any example of using KOS with io? Then I can set all the real parameters for Ferret.

The tests in the kos module should give a sense of how to use the API. Or are you wondering how to use it with an actual network connection, eg TCP?

sinui0:
I'm going to go ahead and merge this into the `ferret` branch and we can continue there

PR #174, opened 2024-07-22
add Lpn estimator

xiangxiecrypto:
This PR contains:
1. A security estimator for LPN in binary field.
2. A security estimator for Dual LPN in binary field.
More work still need to do to implement a "smart" way to choose the "best" parameters according to the number of COTs needed when using Ferret. This will be done later.

### DIFF
--- a/crates/mpz-core/Cargo.toml
+++ b/crates/mpz-core/Cargo.toml
@@ -28,6 +28,7 @@ itybity.workspace = true
 opaque-debug.workspace = true
 bcs = "0.1.5"
 rand_core = "0.6.4"
+rug = "1.24.1"
 bytemuck = { workspace = true, features = ["derive"] }
 generic-array.workspace = true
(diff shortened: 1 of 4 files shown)


sinui0:
@xiangxiecrypto can we convert this into something like:

```rust
/// Selects suitable parameters with a minimum security level of 128 bits.
///
/// # Arguments
/// * `ty` - LPN type.
/// * `k` - Length of the secret vector.
/// * `t` - Hamming weight of the error vector.
fn select_params(ty: LpnType, k: usize, t: usize) -> LpnParameters {}
```

The implementation can be a simple static look up table, or approximate it with a linear regression. Instead of including this in the `mpz-core` library we can convert it into a binary which computes this

themighty1:
Closing, superceded by https://github.com/privacy-scaling-explorations/mpz/pull/218

PR #181, opened 2024-09-05
fix: Use Cargo Clippy to avoid the `gen` reserved keyword

### LINKED ISSUE #180: Replace usage of `gen` prior to next Rust edition
The keyword `gen` is being reserved and we use it various places related to a garbled circuit generator. We must replace our usage.

heeckhau:
RUSTFLAGS="-D keyword_idents_2024" cargo clippy --fix fixes #180


sinui0:
This is a fix, but I would rather rename instead of `r#`. `rand` will be changing `gen` to `random` in the next release. We may want to rename to `Garbler`

sinui0:
Closing, I would prefer we avoid the `r#` solution

PR #171, opened 2024-06-28
add ferret with io

xiangxiecrypto:
see #137 
rebase `dev` and shrink all the commits.

PR #182, opened 2024-09-12
Add QuickSilver with io

xiangxiecrypto:
The current PR implements the QuickSliver protocol with IO.
But still some optimizations are needed.
1. The FS transformation is not `accurate`, the transcripts in the ideal functionalities are not included. 
2. In the protocol, the prover needs to send a bit per AND gate to the verifier, we should optimize it.
3. Heavy operations can be optimized. E.g., computing the powers of `Block` and inner product between two long `Block` vectors.


xiangxiecrypto:
Do you want me to merge #174 #172 #171 to this PR for you to review? @sinui0

sinui0:
Superseded by #198

PR #186, opened 2024-10-23
feat: lazy ot

### LINKED ISSUE #160: Preprocessing Lazy API
With the introduction of the `Allocate` trait in #155 we introduced a "lazy" API for preprocessing OTs. Upon further thought, this pattern can be applied to all offline preprocessing. It could potentially yield significant performance improvements by optimally batching and paralleling workloads.

Let's take a stab at this during the upcoming refactoring of `mpz-garble`

sinui0:
This PR refactors our OT crates to have lazy interfaces. It also disposes of committed/verifiable functionality as we no longer need it. Dependent crates are in a broken state and will be fixed in following PRs. Ferret is also omitted temporarily.


crates/mpz-common/Cargo.toml
```diff
@@ -4,10 +4,14 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = ["sync"]
-sync = ["tokio/sync"]
-test-utils = ["uid-mux/test-utils"]
-ideal = []
+default = []
+ctx = []
```
th4s: In general I am not a fan of fine-grained feature flags because they let the compilation combinations explode.  What is e.g. the benefit of having  `ctx` and `executor` feature flags?

crates/mpz-ot-core/src/cot/derandomize.rs
```diff
@@ -0,0 +1,390 @@
...
+    sender: Sender<COTSenderOutput>,
+}
+
+/// Derandomized COT sender.
+///
+/// This is a COT sender which derandomizes preprocessed RCOTs.
+#[derive(Debug)]
+pub struct DerandCOTSender<T> {
+    rcot: T,
+    pending: usize,
```
th4s: Do we need this field? Isn't it always equal to the length of the inner vectors?

crates/mpz-ot-core/src/cot/derandomize.rs
```diff
@@ -0,0 +1,390 @@
...
+mod tests {
+    use mpz_common::future::Output;
+    use rand::{rngs::StdRng, Rng, SeedableRng};
+
+    use crate::{ideal::rcot::IdealRCOT, test::assert_cot};
+
+    use super::*;
+
+    #[test]
+    fn test_derandomize_cot() {
```
th4s, line 344: Can we have a test which shows the correctness when having several queues and adjustments?

crates/mpz-ot-core/src/ideal/cot.rs
```diff
@@ -1,167 +1,262 @@
...
+    cot::{COTReceiver, COTReceiverOutput, COTSender, COTSenderOutput},
+    TransferId,
+};
+
+type Error = IdealCOTError;
+type Result<T, E = Error> = core::result::Result<T, E>;
+
+#[derive(Debug, Default)]
+struct SenderState {
+    alloc: usize,
```
th4s, line 21: Can we get rid of `alloc`? Doesn't really seem to be used for ideal cot.

crates/mpz-ot-core/src/cot/derandomize.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// Returns `true` if the receiver wants to adjust COTs.
+    pub fn wants_adjust(&self) -> bool {
+        self.pending > 0
+    }
+
+    /// Adjusts the COTs.
+    pub fn adjust(
+        &mut self,
+    ) -> Result<(Derandomize, ReceiveAdjust<'_, T>), DerandCOTReceiverError> {
+        let mut flip = self.derandomize.clone();
```
th4s: Don't we need to consume `self.derandomize` here?  `queue` is already drained in  `ReceiverAdjust`.

crates/mpz-ot-core/src/kos/receiver.rs
```diff
@@ -284,389 +237,125 @@ impl Receiver<state::Extension> {
...
 
-        // Add to existing keys.
-        self.state.keys.extend(keys);
-        self.state.choices.extend(unchecked_choices);
+        // Resolve any queued transfers.
+        if !self.queue.is_empty() {
+            let mut i = 0;
+            for Queued { count, sender } in mem::take(&mut self.queue) {
+                let choices = self.state.choices[i..i + count].to_vec();
+                let msgs = self.state.msgs[i..i + count].to_vec();
```
th4s, line 250: Can't we just
```rust
let choices = self.state.choices.drain(..count)
```
and omit the drains below

crates/mpz-ot/src/cot/derandomize.rs
```diff
@@ -0,0 +1,14 @@
...
+
+pub use receiver::{DerandCOTReceiver, DerandCOTReceiverError};
+pub use sender::{DerandCOTSender, DerandCOTSenderError};
+
+#[cfg(test)]
+mod tests {
+
+    #[tokio::test]
+    async fn test_derandomize_cot() {
+        todo!()
```
th4s: todo

crates/mpz-ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -86,125 +76,137 @@ impl Receiver {
...
+                    |((c, key), [ct0, ct1])| {
+                        if c {
+                            key ^ ct1
+                        } else {
+                            key ^ ct0
+                        }
+                    },
+                );
+
+        while let Some(Queued { count, sender }) = self.queue.pop_front() {
```
themighty1, line 159: if the sender payload length is less than what was queued, then there will be partial or empty `OTReceiverOutput`. Is this the intended behavior? why is such behavior needed?
seems like a footgun which will be hard to debug.

th4s (approved):
lgtm :rocket: nice work

crates/mpz-ot-core/src/cot/derandomize.rs
sinui0: you're correct

crates/mpz-ot-core/src/ideal/cot.rs
sinui0: it ensures both views are synchronized, which will help prevent deadlocks when using real protocols

crates/mpz-ot-core/src/cot/derandomize.rs
sinui0: correct

crates/mpz-ot-core/src/kos/receiver.rs
sinui0: Every call to `drain` involves copying the remaining contents of the vector

themighty1:
LGTM, except one q above @sinui0

sinui0:
> LGTM, except one q above @sinui0 

I dont see a q

themighty1 (commented):
the q should now be seen

crates/mpz-ot-core/src/chou_orlandi/receiver.rs
themighty1: @sinui0 ---^

sinui0: The length of the entire queue gets processed at once, this should always have the correct number. The assumption is that new messages will not be queued in between transfers. This should probably be documented, or enforced via type state.

themighty1: @sinui0 
Sorry, I'll try to more accurately describe the issue because your response didn't seem to address it.

PR #187, opened 2024-10-27
refactor: spawn threads in mt executor

sinui0:
This PR refactors the multithreaded executor to put each logical thread on its own system thread.

# Changes

- `Context` trait updated to have `'static` bounds
- `Spawn` trait exposed so custom thread spawner can be used (for WASM build with webworkers)
- Removed some old macros/functionality that we don't need.


crates/mpz-common/src/executor/mt.rs
```diff
@@ -1,146 +1,219 @@
...
+/// Config for [`MTExecutor`].
+#[derive(Debug, Clone)]
+pub struct MTConfig {
+    max_concurrency: usize,
+}
 
-const MAX_THREADS: usize = 255;
+impl Default for MTConfig {
+    fn default() -> Self {
+        Self { max_concurrency: 8 }
```
th4s, line 38: What about having https://doc.rust-lang.org/std/thread/fn.available_parallelism.html as default?

crates/mpz-common/src/executor/mt.rs
```diff
@@ -150,63 +223,30 @@ where
...
+        &mut self.io
     }
 
     async fn join<'a, A, B, RA, RB>(&'a mut self, a: A, b: B) -> Result<(RA, RB), ContextError>
     where
-        A: for<'b> FnOnce(&'b mut Self) -> ScopedBoxFuture<'a, 'b, RA> + Send + 'a,
-        B: for<'b> FnOnce(&'b mut Self) -> ScopedBoxFuture<'a, 'b, RB> + Send + 'a,
-        RA: Send + 'a,
-        RB: Send + 'a,
+        A: for<'b> FnOnce(&'b mut Self) -> ScopedBoxFuture<'a, 'b, RA> + Send + 'static,
```
th4s, line 235: Can we have `join` and `try_join` for more than 2 tasks?

th4s (approved):
looking very nice and clean :+1:

crates/mpz-common/src/executor/mt.rs
sinui0: This parameter is more about how many times a thread can fork, rather than total parallelism

crates/mpz-common/src/executor/mt.rs
sinui0: Yeah, I'll add this upon needing it

themighty1 (commented):
ACK

PR #188, opened 2024-10-31
feat: lazy ole and share conversion

sinui0:
This PR updates our OLE and share conversion crates to have lazy interfaces. I've also refactored the share conversion core logic a bit, including eliminating a round from A2M.


crates/mpz-fields/src/gf2_128.rs
```diff
@@ -126,14 +139,19 @@ impl Field for Gf2_128 {
     }
 
     /// Galois field inversion of 128-bit block.
-    fn inverse(self) -> Self {
+    fn inverse(self) -> Option<Self> {
+        if self.0 == 0 {
```
th4s: better: `self == Gf2_128::zero()`

crates/mpz-ole-core/src/lib.rs
```diff
@@ -1,146 +1,227 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - Input value, `b`.
+    /// * `masks` - Chosen correlation masks.
+    #[inline]
+    pub(crate) fn new_ole_receiver(
+        input: F,
+        masks: Array<F, F::BitSize>,
+        corr: MaskedCorrelation<F>,
```
th4s, line 121: This input is missing in the function arguments comment.

crates/mpz-ole-core/src/sender.rs
```diff
@@ -1,149 +1,191 @@
...
-        let adjustments = BatchAdjust { id, adjustments };
+            sender.send(ROLESenderOutput {
+                id: self.id.next(),
+                shares,
+            });
+        }
 
-        Some((sender_adjust, adjustments))
-    }
+        self.role.extend_from_slice(&shares[i..]);
```
th4s, line 115: Doesn't this have to be
```rust
self.role.extend_from_slice(&shares);
```

crates/mpz-ole-core/src/sender.rs
```diff
@@ -1,149 +1,191 @@
...
+}
+
+impl<T, F> Sender<T, F>
+where
+    T: ROTSender<[F; 2]>,
+    F: Field,
+{
+    /// Returns `true` if the sender wants to send.
+    pub fn wants_send(&self) -> bool {
+        self.alloc > 0
```
th4s, line 75: hmm shouldn't this be `self.pending > 0`

crates/mpz-ole-core/src/test.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub x: F,
+    pub y: F,
+}
+
+/// Returns a random OLE.
+pub fn role<F: Field, R: Rng + ?Sized>(rng: &mut R) -> OLE<F> {
+    let a = F::rand(rng);
+    let b = F::rand(rng);
+    let x = F::rand(rng);
+    let y = (a * b) - x;
```
th4s, line 25: I think we originally had it like `y = a * b + x`
So with `y = a * b - x`, this would mean that OLE is equivalent to M2A, right?

crates/mpz-ole/src/sender.rs
```diff
@@ -0,0 +1,125 @@
...
+impl<Ctx, T, F> Flush<Ctx> for Sender<T, F>
+where
+    Ctx: Context,
+    T: ROTSender<[F; 2]> + Flush<Ctx> + Send,
+    F: Field + Serialize,
+{
+    type Error = SenderError;
+
+    fn wants_flush(&self) -> bool {
+        self.core.wants_send()
```
th4s, line 70: Why not `self.core.wants_send() || self.core.rot().wants_flush()`

crates/mpz-ole-core/src/lib.rs
```diff
@@ -1,146 +1,227 @@
...
+//! - The sender defines a linear function `y = ab + x` and sends `a` and `x` to
+//!   the functionality.
+//! - The receiver sends their input `b` to the functionality.
+//! - The functionality computes `y = ab + x` and returns `y` to the receiver.
+//!
+//! It's often easier to frame OLE as producing an additive sharing of a
+//! product.
+//!
+//! Where the sender knows `(a, x)` and the receiver knows `(b, y)` such that
+//! `ab = x + y`. This representation is used in [`OLEShare`].
```
th4s, line 15: I think you use `y = a * b - x`, right?

crates/mpz-share-conversion-core/src/a2m.rs
th4s, line 1: lgtm, could not find an attack and maths checks out

th4s (approved):
Nice :rocket: :rocket:

crates/mpz-ole-core/src/sender.rs
sinui0: nope, we want to preprocess based on `alloc`. `pending` is for tracking what has been _queued_

crates/mpz-ole-core/src/sender.rs
sinui0: No this places the _remainder_ of the ROLE into the internal buffer. 0..i has been used to fulfill the queue

crates/mpz-ole-core/src/test.rs
sinui0: Yeah, this makes more sense to have it in this form as we only use it this way

crates/mpz-ole/src/sender.rs
sinui0: I pondered on this for a minute and it seemed to make sense to only flush if the functionality at hand has something to do.

crates/mpz-ole-core/src/lib.rs
sinui0: I introduce the functionality as it is conventionally depicted then in the last paragraph clarify we represent it in this other form.

crates/mpz-share-conversion-core/src/a2m.rs
```diff
@@ -1,124 +1,169 @@
...
+            add: role.add,
+            mul: role.mul,
+        }
     }
 
-    let output = masks.iter().zip(ole_output).map(|(&m, o)| m + o).collect();
-    Ok(output)
+    /// Sends the offset to the sender.
+    pub(crate) fn offset(self) -> (A2MReceiverAdjust<F>, Offset<F>) {
+        let offset = self.input - self.mul;
```
themighty1, line 100: @sinui0 @th4s 
Here, the change is `u = x - x'`  instead of `u = x + x'`. 

While I realize that the tests confirm that the conversion result is correct, Im trying to understand what change need to be done to the original ROLE-OLE from https://crypto.stackexchange.com/questions/100634/converting-a-random-ole-oblivious-linear-function-evaluation-to-an-ole

I couldnt see any other changes in the code to the values parties send and output?
With only this change the math in the SE link does not check out.

themighty1 (commented):
good work, ACK

crates/mpz-share-conversion-core/src/a2m.rs
th4s: The reason for the deviation is, that Alice will keep her original ROLE input, to make the protocol more efficient.
This is because for A2M we only need a random sender input for the OLE.

To make this work, instead of sending `u = x + x'`, Alice now sends `u = x - x'`. This change is necessary, because when Alice wants to keep her original input this means `x = x'` in the adjustment phase. With the original equation she would then send `2x'`, and Bob could recover `x'`. With the adapted equation she would send `u = x' - x' = 0`, which does not need to be sent. It just means Bob keeps his output in the adapted OLE. I think one can consider this as a one-sided adjustment. In order to make the equations work again so that `alpha + beta = xy` we need to flip the equations for `v` and `alpha`, so `v = y - y'` and `alpha = a' + x'v`

themighty1: Thanks for explaining, I see now. All signs need to be flipped except for beta. The math checks out now.

PR #191, opened 2024-11-03
feat: garble vm

sinui0:
This PR implements a semi-honest GC VM, including a lot of the core functionality which can be applied to implement the coming quicksilver VM.


crates/mpz-memory-core/src/correlated/macs.rs
```diff
@@ -0,0 +1,292 @@
...
+
+type Result<T> = core::result::Result<T, MacStoreError>;
+
+/// MAC.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct Mac(Block);
+
+impl Mac {
+    /// Public MACs.
+    pub const PUBLIC: [Mac; 2] = [Mac(MAC_ZERO), Mac(MAC_ONE)];
```
th4s, line 24: What is the idea of public macs? Doesn't this destroy the arithmetic of M_0 + M_1 = DELTA

crates/mpz-garble-core/src/generator.rs
```diff
@@ -361,60 +319,61 @@ where
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use crate::{ChaChaEncoder, Encoder};
-    use mpz_circuits::{circuits::AES128, CircuitBuilder};
-    use pretty_assertions::assert_eq;
+// #[cfg(test)]
```
th4s: todo

crates/mpz-garble-core/src/lib.rs
```diff
@@ -129,119 +89,183 @@ mod tests {
...
+                    .into_iter()
+                    .zip(&output_keys)
+                    .map(|(mac, key)| mac.pointer() ^ key.pointer()),
+            );
+
+            assert_eq!(output, expected);
+        }
     }
+
+    // // Tests garbling a circuit with no AND gates
```
th4s: todo

crates/mpz-garble/src/protocol/semihonest/generator.rs
```diff
@@ -0,0 +1,319 @@
...
+            }
+
+            let outputs = ctx
+                .map(calls, f.clone(), |(call, _)| call.circ().and_count())
+                .await?;
+
+            outputs.into_iter().collect::<Result<(), Error>>()?;
+        }
+
+        self.mark_executed()?;
```
th4s, line 226: Is this the same like in line 274 ?

crates/mpz-garble/src/protocol/semihonest/generator.rs
```diff
@@ -0,0 +1,319 @@
...
+            preprocessed: Vec::new(),
+        }
+    }
+
+    fn take_preprocess_calls(&mut self) -> Vec<(Call, Slice)> {
+        let store = self.store.try_lock().unwrap();
+        self.call_stack
+            .filter_drain(|(call, _)| call.inputs().iter().all(|slice| store.is_set_keys(*slice)))
+            .collect()
+    }
```
th4s, line 45: On the evaluator side you filter for calls which are independent. Why don't you have to do that here?

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
```diff
@@ -0,0 +1,358 @@
...
+                    .inputs()
+                    .iter()
+                    .all(|input| input.to_range().is_disjoint(&idx_outputs))
+                {
+                    idx_outputs |= output.to_range();
+                    true
+                } else {
+                    false
+                }
+            })
```
th4s, line 62: Isn't that dependent on the order of calls in `call_stack`. So don't you have to build `idx_outputs` in a first iteration and then `filter_drain` in a second one?

crates/mpz-garble-core/src/view.rs
```diff
@@ -0,0 +1,467 @@
...
+        let input = range.intersection(&self.input.all);
+        let output = range.intersection(&self.output.all);
+
+        // Transfer decode info.
+        //
+        // Only send decode info for output data and generator's inputs.
+        let decodable_input = match self.role {
+            Role::Generator => input.intersection(self.vis.private()),
+            Role::Evaluator => input.intersection(self.vis.blind()),
+        };
```
th4s, line 285: Why exactly do we need decoding information for input?

th4s (approved):
looking good, nice abstractions :rocket: :rocket:

crates/mpz-memory-core/src/correlated/macs.rs
sinui0: This is used for QS, we don't use it for GC because it prevents us from preprocessing circuits before we know the input visibility.

> Doesn't this destroy the arithmetic of M_0 + M_1 = DELTA

No, the identity of the values they correspond to are distinct. MAC_ZERO is a MAC on a constant 0 value, MAC_ONE is a MAC on a constant 1 value. They share no relation, or put another way they have different keys.

crates/mpz-garble-core/src/view.rs
sinui0: The generator may want to send the decoding information for some of their private inputs at a later point. They could just send them directly, but this just avoids implementing two different flows for inputs vs outputs

crates/mpz-garble/src/protocol/semihonest/generator.rs
sinui0: Very close, but this marks the preprocessed calls as executed. The evaluator executes all preprocessed calls before executing the rest of the callstack.

crates/mpz-garble/src/protocol/semihonest/generator.rs
sinui0: Its effectively the same thing. The input MACs won't be set yet if the calls are dependent. It's asymmetric because the generator knows the output keys after garbling, whereas the evaluator has to actually evaluate the circuit to get the output MACs.

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
sinui0: I don't understand this

th4s: You want to collect calls which have no dependencies on other calls. To do this you check if all inputs of a call are disjoint with `idx_outputs`. But in the same loop `idx_outputs` is built. 

For example you have two calls. Let's name them 1 and 2 and assume call 2 depends on call 1. Now what happens if call 2 is **before** call 1 in `call_stack`. Then the closure will return **true** for call 2, although it depends on call 1. This is because `idx_outputs` does not yet contain the outputs of call 1.

crates/mpz-common/src/executor/mt.rs
```diff
@@ -230,6 +273,66 @@ where
...
+        f: F,
+        weight: W,
+    ) -> Result<Vec<R>, ContextError>
+    where
+        F: for<'b> Fn(&'b mut Self, T) -> ScopedBoxFuture<'static, 'b, R> + Clone + Send + 'static,
+        T: Send + 'static,
+        R: Send + 'static,
+        W: Fn(&T) -> usize + Send + 'static,
+    {
+        let item_count = items.len();
```
themighty1, line 288: Do we want to return an early error here if `item_count` == 0 ?

crates/mpz-common/src/executor/mt.rs
```diff
@@ -200,6 +222,27 @@ impl<Io> MTContext<Io> {
             .first()
             .expect("child thread should be available"))
     }
+
+    async fn get_children(&mut self, count: usize) -> Result<&[Handle<Self>], ContextError> {
```
themighty1, line 226: Is this fn allowed to return more children than `count`? If so, could we add a comment, otherwise such behaviour looks like an error.

crates/mpz-garble-core/src/view.rs
```diff
@@ -0,0 +1,467 @@
...
+
+        self.flush.decode_info |= decodable_input | output.intersection(&self.output.preprocessed);
+
+        // Prove MACs.
+        //
+        // Only prove MACs for output data and evaluator's inputs.
+        let provable_input = match self.role {
+            Role::Generator => input - self.vis.visible(),
```
themighty1, line 293: Is there a reason why this not `input.intersection(self.vis.blind())` like above?

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
```diff
@@ -0,0 +1,358 @@
...
+            .filter_drain(|(call, output)| {
+                if call
+                    .inputs()
+                    .iter()
+                    .all(|input| input.to_range().is_disjoint(&idx_outputs))
+                {
+                    idx_outputs |= output.to_range();
+                    true
+                } else {
+                    false
```
themighty1, line 60: Shouldn't we also do
`                    idx_outputs |= output.to_range();`
before returning `false`?

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
```diff
@@ -0,0 +1,358 @@
...
+    async fn preprocess(&mut self, ctx: &mut Ctx) -> Result<()> {
+        let f = scope_closure(|ctx, (call, output): (Call, Slice)| {
+            async move {
+                let garbled_circuit = receive_garbled_circuit(ctx, call.circ()).await?;
+                Ok::<_, Error>((call, output, garbled_circuit))
+            }
+            .scope_boxed()
+        });
+
+        while !self.call_stack.is_empty() {
```
themighty1, line 229: Since the loop will end up draining the entire callstack, what is the purpose of receiving the garbled circuits layer-by-layer? We can just receive the garbled circuits for the entirety of the callstack in one go.

crates/mpz-memory-core/src/correlated.rs
```diff
@@ -185,10 +204,60 @@ impl BitXor<&Delta> for &Block {
...
+
+        // Commitments must be different.
+        if low == high {
+            return Err(MacCommitmentError {
+                id,
+                kind: MacCommitmentErrorKind::Duplicate,
+            });
+        }
+
+        let expected = hasher.tccr(Block::from((id as u128).to_be_bytes()), *mac.as_block());
```
themighty1, line 232: I think we should use a strong hash here. Fixed-key AES TCCR was not proven to have strong hash properties like e.g. collision resistance.

crates/mpz-memory-core/src/correlated/keys.rs
```diff
@@ -0,0 +1,408 @@
...
+impl Key {
+    /// Returns the pointer bit.
+    #[inline]
+    pub fn pointer(&self) -> bool {
+        self.0.lsb()
+    }
+
+    /// Adjusts the truth value of the corresponding MAC.
+    #[inline]
+    pub fn adjust(&mut self, adjust: bool, delta: &Delta) {
```
themighty1, line 33: I had a hard time parsing what this method tries to achieve. Could you describe the `adjust` argument?

crates/mpz-memory-core/src/correlated/keys.rs
```diff
@@ -0,0 +1,408 @@
...
+    public_one: Key,
+    used: RangeSet,
+}
+
+impl KeyStore {
+    /// Creates a new key store.
+    #[inline]
+    pub fn new(delta: Delta) -> Self {
+        let mut public_one = Key(MAC_ONE ^ delta.as_block());
+        public_one.0.set_lsb(false);
```
themighty1, line 161: iiuc, at this point the LSB will always be 0, why do we need to set it here?
`MAC_ONE`'s LSB is always 1, `delta`'s LSB is always 1 and so the Key's LSB will always be zero.

crates/mpz-memory-core/src/correlated/keys.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Allocates uninitialized memory.
+    #[inline]
+    pub fn alloc(&mut self, len: usize) -> Slice {
+        self.keys.alloc(len)
+    }
+
+    /// Allocates memory with the given keys.
+    ///
+    /// The provided keys are marked as used.
```
themighty1, line 202: The keys are not marked as used here.

themighty1 (commented):
Good work. 
I left some qs, including security-critical ones.
@sinui0

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
th4s: I agree, makes sense and I tested it. Without this fix the vm preprocess call hangs. With this fix it proceeds.

sinui0: Please open a PR if you've identified a fix for a bug

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
sinui0: It's because we process them in parallel and this ensures no interdependencies to facilitate that. The generator is streaming the gates over separate IO threads

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
th4s: included in https://github.com/privacy-scaling-explorations/mpz/pull/197

crates/mpz-common/src/executor/mt.rs
themighty1: doesnt seem necessary

crates/mpz-common/src/executor/mt.rs
themighty1: N/A in the latest refactor

crates/mpz-memory-core/src/correlated.rs
themighty1: https://github.com/privacy-scaling-explorations/mpz/issues/260

crates/mpz-memory-core/src/correlated/keys.rs
themighty1: a comment is now present in that fn

crates/mpz-memory-core/src/correlated/keys.rs
themighty1: i assume it's for expliciteness, ill add a comment.

crates/mpz-memory-core/src/correlated/keys.rs
themighty1: stray comment

sinui0: Fixed in https://github.com/privacy-scaling-explorations/mpz/commit/d850507c9dc8b9a776302f55b32324b2f84468da

PR #196, opened 2024-12-12
chore: cargo clippy fixes for 1.83.0

heeckhau:


crates/mpz-core/src/lpn.rs
```diff
@@ -64,7 +64,7 @@ impl<const D: usize> LpnEncoder<D> {
 
     #[inline]
     fn compute_one_row(&self, y: &mut [Block], x: &[Block], pos: usize, prp: &Prp) {
-        let block_size = (D + 4 - 1) / 4;
+        let block_size = D.div_ceil(4);
```
th4s, line 67: I am not sure if it is helpful here to follow clippy's advice. I suspect that this can create confusion so I would tend to suppress clippy or add a comment.

crates/mpz-ot-core/src/ferret/mpcot/receiver_regular.rs
```diff
@@ -47,7 +47,7 @@ impl Receiver<state::PreExtension> {
         }
 
         // The range of each interval.
-        let k = (n + t - 1) / t;
+        let k = n.div_ceil(t);
```
th4s, line 50: same

crates/mpz-ot-core/src/ferret/mpcot/sender_regular.rs
```diff
@@ -50,7 +50,7 @@ impl Sender<state::PreExtension> {
         }
 
         // The range of each interval.
-        let k = (n + t - 1) / t;
+        let k = n.div_ceil(t);
```
th4s, line 53: same

th4s (approved):

crates/mpz-core/src/lpn.rs
heeckhau: I assume the Clippy team made this rule with severity `error` for a reason.
The docs says: **It's simpler, clearer and more readable.**

https://rust-lang.github.io/rust-clippy/master/index.html#manual_div_ceil
https://github.com/rust-lang/rust-clippy/issues/12894

th4s: I don't have a strong opinion here. My thoughts were that I would probably spend a few extra brain cycles what happened to the original formula.

PR #197, opened 2024-12-12
Fix deadlock for garble-vm

th4s:
This PR fixes the VM hanging, when circuit output ranges are used as input ranges for a new circuit. **The change fixes the curently hanging tests of the prf and cipher crates of tlsnotary.**

When a circuit is executed the input range corresponding to the output range of the circuit output is not marked as `complete`. This leads to `is_committed` returning false and the circuit never being executed.

**Questions**
1) Do we also need to touch `assigned` or `all` of `InputView`? Anything else which needs to be adapted for consistency?
2) Is this change compatible with the overall picture, because I get the idea that so far inputs and outputs are treated entirely different.

### DIFF
--- a/crates/mpz-garble-core/src/view.rs
+++ b/crates/mpz-garble-core/src/view.rs
@@ -207,10 +207,6 @@ impl View {
 
         self.output.preprocessed |= &range;
         self.output.complete |= &range;
-
-        // If we want to use the output as input for another circuit
-        self.input.complete |= &range;
-
         // If marked for decoding, transfer decode info.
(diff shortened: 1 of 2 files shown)


sinui0:
This isn't the right solution, as we do need to distinguish between input and output values. The right approach is likely to modify `is_committed` so that it considers completed outputs as committed.

th4s:
Adapted `is_committed` now instead of mixing input and output ranges. I also included Dan's suggestion to fix the preprocessing in this PR.

themighty1:
ack

PR #198, opened 2024-12-13
feat: quicksilver

sinui0:
This PR provides a zkVM implementation using Quicksilver. Builds on work from @xiangxiecrypto and supercedes #172 and #182.

### DIFF
--- a/crates/mpz-zk-core/src/check.rs
+++ b/crates/mpz-zk-core/src/check.rs
@@ -104,7 +104,7 @@ impl Check {
                     .map(|(macs, chi)| compute_terms(macs, chi))
                     .reduce(
                         || (Block::ZERO, Block::ZERO),
-                        |(u_0, v_0), (u_1, v_1)| (u_0 ^ u_1, v_0 ^ v_1),
+                        |(u_acc, v_acc), (u, v)| (u_acc ^ u, v_acc ^ v),
                     );
             } else {
                 let (mut u, mut v) = macs
@@ -113,7 +113,7 @@ impl Check {
                     .map(|(macs, chi)| compute_terms(macs, chi))
                     .fold(
                         (Block::ZERO, Block::ZERO),
-                        |(u_0, v_0), (u_1, v_1)| (u_0 ^ u_1, v_0 ^ v_1),
+                        |(u_acc, v_acc), (u, v)| (u_acc ^ u, v_acc ^ v),
                     );
             }
         }
@@ -162,7 +162,7 @@ impl Check {
                     .map(|(keys, chi)| compute_term(keys, chi, delta))
                     .reduce(
                         || Block::ZERO,
-                        |w_0, w_1| w_0 ^ w_1,
+                        |w_acc, w| w_acc ^ w,
                     );
             } else {
                 let mut w = keys
@@ -171,7 +171,7 @@ impl Check {
                     .map(|(keys, chi)| compute_term(keys, chi, delta))
                     .fold(
                         Block::ZERO,
-                        |w_0, w_1| w_0 ^ w_1,
+                        |w_acc, w| w_acc ^ w,
                     );
             }
         }
--- a/crates/mpz-zk-core/src/check.rs
+++ b/crates/mpz-zk-core/src/check.rs
@@ -1,0 +1,204 @@
+//! Quicksilver consistency check.
+
+use std::mem;
+
+use blake3::Hasher;
+use cfg_if::cfg_if;
+use mpz_core::{
+    bitvec::{BitSlice, BitVec},
+    Block,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::vole::{vole_receiver, vole_sender};
+
+type Result<T> = core::result::Result<T, CheckError>;
+
+/// Values sent from the prover to the verifier for the consistency check.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct UV {
+    u: Block,
+    v: Block,
+}
+
+#[derive(Debug, Default, Clone, Copy)]
+pub(crate) struct Triple {
+    pub(crate) x: Block,
+    pub(crate) y: Block,
+    pub(crate) z: Block,
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct Check {
+    transcript: Hasher,
+    triples: Vec<Triple>,
+    adjust: BitVec<u8>,
+}
+
+impl Check {
+    /// Reserves capacity for at least `n` AND gates, returns the starting
+    /// index.
+    pub(crate) fn reserve(&mut self, n: usize) -> usize {
+        let idx = self.triples.len();
+        self.triples.resize_with(idx + n, Default::default);
+        self.adjust.resize_with(idx + n, |_| Default::default());
+        idx
+    }
+
+    pub(crate) fn write(&mut self, idx: usize, triples: &[Triple], adjust: &BitSlice<u8>) {
+        self.triples[idx..idx + triples.len()].copy_from_slice(triples);
+        self.adjust[idx..idx + triples.len()].copy_from_bitslice(adjust);
+    }
+
+    /// Returns `true` if there are gates to check.
+    #[inline]
+    pub(crate) fn wants_check(&self) -> bool {
+        !self.triples.is_empty()
+    }
+
+    fn compute_chis(&self) -> Vec<Block> {
+        // TODO: Consider using a PRG instead so computing the coefficients
+        // can be done in parallel.
+        let mut chi = Block::try_from(&self.transcript.finalize().as_bytes()[..16])
+            .expect("block should be 16 bytes");
+        let mut chis = Vec::with_capacity(self.triples.len());
+        chis.push(chi);
+        for _ in 1..self.triples.len() {
+            chi = chi.gfmul(chi);
+            chis.push(chi);
+        }
+
+        chis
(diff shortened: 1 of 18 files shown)


crates/mpz-zk-core/src/check.rs
```diff
@@ -0,0 +1,204 @@
...
+            if #[cfg(all(feature = "rayon", not(feature = "force-st")))] {
+                use rayon::prelude::*;
+
+                let (mut u, mut v) = macs
+                    .into_par_iter()
+                    .zip(chis)
+                    .map(|(macs, chi)| compute_terms(macs, chi))
+                    .reduce(
+                        || (Block::ZERO, Block::ZERO),
+                        |(u_0, v_0), (u_1, v_1)| (u_0 ^ u_1, v_0 ^ v_1),
```
th4s: can we rename `(u_0, v_0)` to `(u_acc, v_acc)`, and `(u_1, v_1)` to `(u, v)`, because I think that conveys better what is going on here.

crates/mpz-zk-core/src/check.rs
```diff
@@ -0,0 +1,204 @@
...
+            if #[cfg(all(feature = "rayon", not(feature = "force-st")))] {
+                use rayon::prelude::*;
+
+                let mut w = keys
+                    .into_par_iter()
+                    .zip(chis)
+                    .map(|(keys, chi)| compute_term(keys, chi, delta))
+                    .reduce(
+                        || Block::ZERO,
+                        |w_0, w_1| w_0 ^ w_1,
```
th4s: same

th4s (approved):
Very nice :+1:, looking pretty clean. Not much to add.

crates/mpz-zk-core/src/vole.rs
```diff
@@ -0,0 +1,65 @@
+//! Vector oblivious polynomial evaluation (VOPE) protocol for degree 1.
+//!
+//! This implementation is specifically for degree 1, i.e. VOLE over the
+//! extension field.
```
xiangxiecrypto, line 4: I think you meant VOPE in this file? 
This implementation simplies figure 4 for a polynomial with degree 1.

xiangxiecrypto (commented):
I do not have too much context of the vm design here. But I checked the impl. of the protocol itself, and it looks pretty good to me.

Just one more thing. In the current implementation, check always occurs at the end of executing the circuit?

sinui0:
> Just one more thing. In the current implementation, check always occurs at the end of executing the circuit?

Yes, the consistency check is always performed after executing the entire callstack. It batches the checks of all circuits together

crates/mpz-zk-core/src/vole.rs
sinui0: Isn't degree 1 VOPE equivalent to VOLE?

xiangxiecrypto: It is different. For boolean circuits, VOLE is Random COT. VOLE is a generalized version to extend boolean to arithmetic circuits.
VOPE focuses on polynomial relations. B = A_0 + A_1*Delta +..., but all the B and A_i are in F_2^{128}.

xiangxiecrypto:
> > Just one more thing. In the current implementation, check always occurs at the end of executing the circuit?
> 
> Yes, the consistency check is always performed after executing the entire callstack. It batches the checks of all circuits together

So that means if the circuit is huge, we have to handle very long vectors of F_2^{128} for the check, which may cause memory issues?

crates/mpz-zk-core/src/vole.rs
sinui0: My understanding is that VOLE is exactly equivalent to VOPE for degree 1, ie a linear function. All the terms in a VOLE are of the same field $\mathbb{F}_{2^{128}}$. The "boolean" variant is often referred to as subfield VOLE (or COT) where `A_1` is in $\mathbb{F}_2$.

If I'm not mistaken, the algorithm used here is the typical way to construct a VOLE from sVOLEs. The QS paper generalizes it for proving higher degrees but we won't be needing that

sinui0:
> So that means if the circuit is huge, we have to handle very long vectors of F_2^{128} for the check, which may cause memory issues?

This is true, and something which we may need to address by introducing a buffer limit which triggers a check. There is a round complexity vs memory trade off there. Shouldn't be an issue and implementing said limit is fairly simple

xiangxiecrypto:
> > So that means if the circuit is huge, we have to handle very long vectors of F_2^{128} for the check, which may cause memory issues?
> 
> This is true, and something which we may need to address by introducing a buffer limit which triggers a check. There is a round complexity vs memory trade off there. Shouldn't be an issue and implementing said limit is fairly simple

Ok, the original version uses a buffer to handle this. I think it wont increase round complexity, since there is only one round. But anyway it is not a big issue.

xiangxiecrypto (approved):

sinui0:
> Ok, the original version uses a buffer to handle this. I think it wont increase round complexity, since there is only one round. But anyway it is not a big issue.

Good point, I forgot that it is only 1 message from P to V. I was focused on batching multiple circuits together and lost that detail.

maceip:
legendary cl

crates/mpz-zk-core/src/view.rs
```diff
@@ -0,0 +1,428 @@
...
+        // Ignore already decoded data.
+        let undecoded = range.difference(&self.decode.complete);
+        if undecoded.is_empty() {
+            return Ok(());
+        }
+
+        self.decode.all |= &undecoded;
+
+        let input = range.intersection(&self.input.all);
+        let output = range.intersection(&self.output.all);
```
themighty1, line 246: I think we want `undecoded.intersection()` on these lines

crates/mpz-zk/src/prover.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    fn mark_private_raw(&mut self, slice: Slice) -> Result<()> {
+        self.store.mark_private_raw(slice)?;
+
+        self.ot.alloc(slice.len()).map_err(Error::ot)?;
+
+        Ok(())
+    }
+
+    fn mark_blind_raw(&mut self, slice: Slice) -> Result<()> {
```
themighty1, line 237: Since there are no blind values for a zk prover, this should return an error.

crates/mpz-zk/src/prover.rs
```diff
@@ -0,0 +1,291 @@
...
+            let masks = BitVec::from_iter(masks);
+            let macs = Mac::from_blocks(macs);
+
+            self.store.set_macs(&masks, &macs)?;
+        }
+
+        while self.store.wants_flush() {
+            let flush = self.store.send_flush()?;
+            ctx.io_mut().send(flush).await?;
+            let flush = ctx.io_mut().expect_next().await?;
```
themighty1, line 68: will `await`ing on the verifier's flush add to the prover's latency? Are there ways to avoid this?

crates/mpz-zk-core/src/store.rs
```diff
@@ -0,0 +1,107 @@
...
+pub use verifier::{VerifierStore, VerifierStoreError};
+
+use blake3::Hash;
+use mpz_core::bitvec::BitVec;
+use serde::{Deserialize, Serialize};
+
+use crate::view::FlushView;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ProverFlush {
```
themighty1, line 14: Do we want `ProverFlushUnchecked` to enforce `adjust.len() == view.commit().len()` ?

crates/mpz-zk/src/verifier.rs
```diff
@@ -0,0 +1,281 @@
...
+where
+    OT: RCOTSender<Block>,
+{
+    type Error = Error;
+
+    fn mark_public_raw(&mut self, slice: Slice) -> Result<()> {
+        self.store.mark_public_raw(slice).map_err(Error::from)
+    }
+
+    fn mark_private_raw(&mut self, slice: Slice) -> Result<()> {
```
themighty1, line 220: this should return an error since there are no private values for a zk verifier

themighty1 (commented):
Good job, there are some questions which need to be addressed.

themighty1:
tagging @sinui0 to look at the comments (i'll remind you later, so this review won't get lost)

crates/mpz-zk/src/prover.rs
themighty1: opened an issue https://github.com/privacy-scaling-explorations/mpz/issues/245

crates/mpz-zk-core/src/view.rs
themighty1: https://github.com/privacy-scaling-explorations/mpz/pull/261

crates/mpz-zk/src/prover.rs
themighty1: https://github.com/privacy-scaling-explorations/mpz/pull/261

crates/mpz-zk-core/src/store.rs
themighty1: https://github.com/privacy-scaling-explorations/mpz/pull/261

crates/mpz-zk/src/verifier.rs
themighty1: https://github.com/privacy-scaling-explorations/mpz/pull/261

Issue #194, opened 2024-12-03
bug(ot): co15 doesnt enforce correct ot count


sinui0:
Suppose, the receiver queues 60 OTs.
Suppose in `receive` the length of `SenderPayload` is 30.

Why is it not an error that the receiver will get only 30 OTs out of the 60 that they intended to receive?

_Originally posted by @themighty1 in https://github.com/privacy-scaling-explorations/mpz/pull/186#discussion_r1867807280_

n-lebel:
Is there currently an error type which can be sent between players to act as an abort? If there was, an easy way to solve this would be for the Sender to check the `ReceiverPayload` length and send a `Result<SenderPayload, Abort>` as its next message instead of just `SenderPayload`. None of the OT error types currently implement `Serialize/Deserialize` so it's not immediately clear to me how to integrate this without adding a type.

n-lebel:
Something along the lines of https://github.com/privacy-scaling-explorations/mpz/pull/199 might help fix the issue. There's probably a better way of organizing the errors, definitely open to suggestions.

sinui0:
Hi @n-lebel thanks for the attention on the issue. I should have specified that this issue pertains to the `alpha.1` branch, where a lot of changes are being aggregated.

Regarding the solution, we specifically avoid returning error information to peers as doing so is a known avenue for attacks. Returning an error locally is sufficient for debugging purposes.

n-lebel:
Good to know, I'll have a look at that branch in that case. Just to make sure about the attack vector comment, you're talking about communicating errors more generally and not this specific solution (assuming everything is constant-time), correct?

sinui0:
Yes, speaking in general we don't send errors over the wire

PR #199, opened 2024-12-28
Fix CO15 mismatched length issue

n-lebel:
This PR adds a new serializable `AbortError` type meant to be sent between players to signal an abort during the protocol. In this case, an abort can be sent by the Sender if the `ReceiverPayload` length does not match their locally requested OT batch size. 

Generally speaking, it may make sense to consider generalizing the use of `AbortError` as for example, a local error by an OT protocol is currently only propagated locally and not to the other player.


th4s:
Thanks for the PR, reading #194 my understanding is that this PR can be closed? Is this correct?

n-lebel:
Yes that's correct, thanks!

PR #202, opened 2025-01-20
feat: ferret

sinui0:
This PR refactors the existing Ferret implementation to fit into the recent interface changes. I've also done some cleaning up here and there, including removing a quadratic cost algorithm present in the uniform MPCOT implementation.

Builds on #171


crates/mpz-core/src/ggm.rs
```diff
@@ -0,0 +1,232 @@
...
+    leaves: &'a mut [Block],
+}
+
+impl<'a> GgmTree<'a> {
+    /// Creates a new GGM tree.
+    ///
+    /// # Arguments
+    ///
+    /// * `depth` - The depth of the tree.
+    /// * `seed` - The seed of the tree.
```
th4s, line 37:
```suggestion
    /// * `seed` - The seed of the tree.
    /// * `leaves` - The last layer of blocks.
```

crates/mpz-core/src/ggm.rs
```diff
@@ -0,0 +1,232 @@
...
+        &self.buf[0]
+    }
+
+    /// Returns the depth of the tree.
+    pub fn depth(&self) -> usize {
+        self.depth
+    }
+
+    /// Returns the layer at the given height.
+    pub fn layer(&self, height: usize) -> Option<&[Block]> {
```
th4s: Isn't height the same as depth?

crates/mpz-core/src/ggm.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// # Panics
+    ///
+    /// - If the position is out of bounds.
+    /// - If the length of the sums is not equal to the depth minus one.
+    ///
+    /// # Arguments
+    ///
+    /// * `depth` - Depth of the tree.
+    /// * `sums` - Sum of the left or right nodes for each layer.
+    /// * `idx` - Index of the missing leaf.
```
th4s, line 74: same

crates/mpz-core/src/ggm.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Returns an iterator over the layers of the GGM tree.
+    pub fn iter_layers(&self) -> impl Iterator<Item = &[Block]> {
+        (0..=self.depth).flat_map(|i| self.layer(i))
+    }
+
+    /// Returns the sums of the left and right nodes for each layer.
+    pub fn layer_sums(&self) -> impl Iterator<Item = [Block; 2]> + '_ {
+        self.iter_layers().skip(1).map(|layer| {
+            let mut left = Block::ZERO;
+            let mut right = Block::ZERO;
```
th4s, line 160: Why do we have to set both to zero here?

crates/mpz-ot-core/src/ferret/config.rs
```diff
@@ -0,0 +1,188 @@
...
+    LpnParameters {
+        n: 5_324_800,
+        k: 240_000,
+        t: 1_300,
+    },
+    LpnParameters {
+        n: 10_485_760,
+        k: 458_000,
+        t: 1280,
+    },
```
th4s, line 187: Are these numbers from the Ferret paper? I cannot find them.

crates/mpz-ot-core/src/ferret/cuckoo.rs
```diff
@@ -0,0 +1,192 @@
+//! Implementation of Cuckoo hash.
+
+use std::array::from_fn;
+
+use mpz_core::{aes::AesEncryptor, Block};
+
+pub(crate) const HASH_NUM: u32 = 3;
+const TRIAL_NUM: usize = 100;
```
th4s, line 8: Is there a reason for why `100`?

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,249 @@
...
+
+                // Generates queries for SPCOT.
+                // See Step 4 in Figure 7.
+                let mut idxs = vec![];
+                let mut spcot_log2_lengths = vec![];
+                let mut spcot_lengths = vec![];
+                for (item, bucket_length) in cuckoo.iter().zip(buckets.iter_buckets()) {
+                    // pad to power of 2.
+                    let power_of_two = (bucket_length + 1)
+                        .checked_next_power_of_two()
```
th4s, line 75: I cannot find this rounding up to the next power of 2 in the paper. On the other hand the SPCOT protocol only works with powers of 2. Is this somehow implied or how did you come up with it?

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -0,0 +1,249 @@
...
+                        .expect("bucket length should be less than usize::MAX / 2 - 1");
+
+                    if let Some(x) = item {
+                        let (_, pos) = buckets.get(x.value as usize)[x.hash_idx as usize];
+
+                        idxs.push(pos);
+                    } else {
+                        idxs.push(bucket_length);
```
th4s, line 83: Shouldn't this be `bucket_length + 1`?

crates/mpz-ot-core/src/ferret/sender.rs
```diff
@@ -0,0 +1,425 @@
...
+        });
+
+        Ok(())
+    }
+
+    /// Performs extension.
+    ///
+    /// # Arguments
+    ///
+    /// * `derandomize` - SPCOT derandomization bits from the receiver.
```
th4s:
```suggestion
    /// * `msg` - SPCOT derandomization bits from the receiver.
```

crates/mpz-ot-core/src/ferret/sender.rs
```diff
@@ -0,0 +1,425 @@
...
+        });
+
+        Ok(SenderExtend { ms, sums })
+    }
+
+    /// Performs the SPCOT consistency check.
+    ///
+    /// # Arguments
+    ///
+    /// * `derandomize` - SPCOT derandomization bits from the receiver.
```
th4s: same

crates/mpz-ot-core/src/ferret/receiver.rs
```diff
@@ -0,0 +1,439 @@
...
+        let params = self.config.select_params(self.macs.len(), self.alloc);
+
+        let err = sample_error_indices(&mut self.prg, lpn_type, params.n, params.t);
+
+        let (mpcot, spcot_lengths, spcot_idxs) =
+            MPCOTReceiver::new(public_prg.gen(), lpn_type).start_extend(&err, params.n)?;
+
+        let spcot_count: usize = spcot_lengths.iter().sum();
+        let masks = &self.choices[self.choices.len() - spcot_count..];
+        let derandomize = self.spcot.derandomize(&spcot_lengths, &spcot_idxs, masks)?;
```
th4s, line 148: What about truncating `self.choices` here, instead of line 193?

th4s (approved):
:rocket: looking pretty good! Some comments, mostly questions.

crates/mpz-ot-core/src/ferret/config.rs
```diff
@@ -0,0 +1,188 @@
...
+fn iteration_cost(ty: LpnType, params: LpnParameters) -> usize {
+    match ty {
+        // The number here is a rough estimation to ensure sufficient buffer.
+        // It is hard to precisely compute the number because of the Cuckoo hashes.
+        LpnType::Uniform => {
+            let m = (1.5 * (params.t as f32)).ceil() as usize;
+            m * ((2 * HASH_NUM as usize * params.n / m)
+                .checked_next_power_of_two()
+                .expect("The length should be less than usize::MAX / 2 - 1")
+                .ilog2() as usize)
```
th4s, line 132: Can you explain how you arrive at this specific part?

crates/mpz-ot-core/src/ferret/config.rs
```diff
@@ -0,0 +1,188 @@
...
+fn default_parameter_selector(ty: LpnType, available: usize, additional: usize) -> LpnParameters {
+    let params = match ty {
+        LpnType::Uniform => UNIFORM_PARAMS,
+        LpnType::Regular => REGULAR_PARAMS,
+    };
+
+    // *Assumes the parameters are in ascending order.*
+    for param in params {
+        let cost = iteration_cost(ty, *param);
+        let net = param.t - cost;
```
th4s, line 107: Why is `net = t - cost`?

crates/mpz-core/src/ggm.rs
sinui0: Because we expanded the PRG in the previous layer using an empty seed, we have to zero it. We could avoid that PRG expansion but it just adds more complexity and probably branching cost.

crates/mpz-ot-core/src/ferret/config.rs
sinui0: These came from Xiang's PR. We do need to confirm/replace them with the work on the parameter table.

crates/mpz-ot-core/src/ferret/cuckoo.rs
sinui0: Xiang's selection. It just puts an upper bound on recursion in the worst case.

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
sinui0: Again, a simplification from Xiang's impl. The extra length in the MPCOT gets discarded.

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
sinui0: Indeed the paper specifies that, which I don't understand as that is outside the bound of the SPCOT vector. It also says on Page 9 sec. 4 that $p_j$ can be set to any value in this case.

crates/mpz-ot-core/src/ferret/config.rs
sinui0: I can't, no. It's from Xiang's PR. It's a rough estimate given that the cuckoo hashing bucket lengths will vary.

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
sinui0: @xiangxiecrypto are you able to explain this?

crates/mpz-ot-core/src/ferret/config.rs
sinui0: `t` is how many COTs will be produced from the iteration. We subtract the cost to execute that iteration to get the net new COTs.

th4s: My understanding is that `n` is the number of COTs that will be produced. I think `t` is the number of LPN errors (hamming weight of the error vector). So I think this should be `net = param.n - cost`, no?

themighty1: seems like a bug, i opened an issue https://github.com/privacy-scaling-explorations/mpz/issues/248

themighty1:
lgtm, opened PR with improvements and also a bug https://github.com/privacy-scaling-explorations/mpz/issues/248

PR #203, opened 2025-01-20
fix(garble): execute entire preprocessed callstack

sinui0:
This PR fixes a bug in the garble VMs preprocessing control flow. It now iteratively executes *all* preprocessed calls instead of just one layer.


th4s:
looking good, unfortunately the integration test still hangs with preprocessing enabled, even with this fix.

themighty1:
ack

PR #204, opened 2025-01-22
refactor: add Callable trait and Vm supertrait

sinui0:
This PR renames `Vm` to `Callable` and adds a `Vm` trait which is a supertrait of the others. This makes bounds cleaner in consuming code.


th4s:
:+1:

PR #206, opened 2025-01-24
refactor: concrete Context type

sinui0:
This PR refactors the `Context` abstraction to be a concrete type.

Not particularly polished, but sufficient for now.

# Changes

- `Context` is now a struct an dynamically dispatches between either single threaded or multi threaded execution.
- `Io` is now a struct, and dynamically dispatches between either a memory channel or a framed byte transport. For now I've coupled the framed transport to use `bincode` but later we can open this up if there is reason to.
- Added ability to construct the multithreaded context using a custom thread spawner callback, and also the ability to provide a multiplexer implementing traits from `uid-mux`.


th4s (approved):

themighty1:
ack

PR #213, opened 2025-01-31
feat(vm): more execute methods

sinui0:
This PR adds additional control flow methods to the `Execute` trait.

### DIFF
--- a/crates/mpz-garble/src/protocol/semihonest/evaluator.rs
+++ b/crates/mpz-garble/src/protocol/semihonest/evaluator.rs
@@ -208,6 +208,10 @@ where
     COT: COTReceiver<bool, Block> + Flush + Send + 'static,
     COT::Future: Send + 'static,
 {
+    fn wants_flush(&self) -> bool {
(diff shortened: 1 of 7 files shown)


th4s (commented):
:+1:

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
```diff
@@ -217,6 +221,18 @@ where
         Ok(())
     }
 
+    fn wants_preprocess(&self) -> bool {
```
themighty1, line 224: Will there be a situation where the first call in the call_stack is not ready for preprocessing?
Otherwise, this fn can just be
```!self.call_stack.is_empty()```

themighty1 (commented):
lgtm, one q for @sinui0

crates/mpz-garble/src/protocol/semihonest/evaluator.rs
themighty1: opened https://github.com/privacy-scaling-explorations/mpz/issues/277

PR #218, opened 2025-02-04
Lpn Estimator with Lookup

th4s:
This PR builds on the work of @xiangxiecrypto and:

- moves the estimator into its own (excluded) workspace crate
- improves the performance of `security_under_agb_binary` a little
- turns the examples into binaries
- adds a `scan` function which samples the parameter space with fixed `t`
- adds code generation of suited lpn instances into static arrays for `ot-core/ferret/config`
- supports verification of lookup tables against a reference implementation


th4s:
@sinui0 I think you can take a look. Now `bin/estimator.rs` does code generation.

crates/lpn-estimator/lpnestimator.com/estimator.py
sinui0, line 1: Let's not check in this code. Link to it in the README if you think its a good resource

crates/lpn-estimator/src/bin/lpn_checker.rs
sinui0, line 1: I don't think this is necessary to check in. You've already validated that the estimator in the PR is more conservative than the reference. I don't want to add that code as a dependency

sinui0 (changes requested):
Good stuff

We will want to have more than 2 parameter options. The original motivation was to have a lookup table so that we can select the best parameters to compute `n` COTs. A reasonable start would be to have an entry for $n = 2^i, i \in 16..24$

crates/lpn-estimator/src/bin/lpn_checker.rs
th4s: Well, I validated it on a subset of the parameter space. But I cannot guarantee that this holds true for the whole parameter space. Also it is impossible to check the whole parameter space because of the abysmal performance of the python version.

So I think it is either check-in the python code and being able to do cross checks or less dependency but trust the rust implementation.

crates/lpn-estimator/lpnestimator.com/estimator.py
th4s: discussion below

th4s:
I changed it so that it produces more output, coming very close to what you proposed, with i= {18..25} since the estimator does not work well for very small numbers. We also have to respect that (n/t) = 2^c for some c.

th4s:
- added tables which produced very good bench results
- removed `lpn_checker` and forked the estimator

sinui0 (approved):

PR #222, opened 2025-02-10
feat(common): try_join variants

sinui0:
This PR provides more `try_join` variants for 3 and 4 closures, similarly to the `futures` crate. This may not survive longer term as part of the API but it's needed now. In the future we probably want to figure out a join scope abstraction.


th4s (commented):
:+1:

PR #224, opened 2025-02-12
refactor: odds and ends

sinui0:
This PR makes a couple small changes:

1. Switches `tlsn-utils` to `dev`
2. Fixes getter impl on VMs to return `None` instead of an error.
3. Fixes a bug in the garble memory implementation which caused duplicate encoding operations.


th4s (commented):
:+1:

PR #221, opened 2025-02-10
Add `get_macs` and `get_keys` methods for ZK Prover/Verifier

th4s:
~~This PR adds an `Encodings` trait to pull out macs and keys from the vm.~~
Add methods for getting keys and macs on ZK prover and verifier

PR #229, opened 2025-02-14
Add Memory trait for Encodings

th4s:
This PR implements the `Memory` trait to deal with raw encodings.


th4s:
Will be replaced by #221.

PR #230, opened 2025-02-14
Fix compile warnings

th4s:
This PR fixes warnings during compilation.

### DIFF
--- a/crates/mpz-garble-core/Cargo.toml
+++ b/crates/mpz-garble-core/Cargo.toml
@@ -9,9 +9,6 @@ workspace = true
 [lib]
 name = "mpz_garble_core"
 
-[profile.release]
-lto = true
-
 [features]
 default = ["rayon"]
 rayon = ["dep:rayon"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ resolver = "2"
 
 [workspace.lints.rust]
 # unsafe_code = "forbid"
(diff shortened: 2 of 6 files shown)


crates/mpz-garble-core/Cargo.toml
```diff
@@ -9,9 +9,6 @@ workspace = true
 [lib]
 name = "mpz_garble_core"
 
-[profile.release]
-lto = true
-
```
th4s, line 14: Nor sure what was the reason here, so pointing at this specifically. I removed it since it is ignored anyways when building the workspace. I assumed we do not want full LTO support on the whole workspace?

sinui0: Not sure why you want to remove this? I don't see a compiler warning

th4s: I am getting this when doing `cargo check`
```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/th4s/tlsnotary/mpz/crates/mpz-garble-core/Cargo.toml
workspace: /home/th4s/tlsnotary/mpz/Cargo.toml
```

sinui0: Hmm ok, let's add LTO to the workspace release profile then. It can yield good performance improvements.

sinui0 (approved):

PR #231, opened 2025-02-14
Fix clippy lints

th4s:
This PR fixes clippy lints.

### DIFF
--- a/crates/mpz-core/src/ggm.rs
+++ b/crates/mpz-core/src/ggm.rs
@@ -1,5 +1,6 @@
 //! GGM tree.
 
+use std::cmp::Ordering;
 use std::ops::Range;
 
 use itybity::ToBits;
@@ -139,13 +140,11 @@ impl<'a> GgmTree<'a> {
 
     /// Returns the layer at the given depth.
     pub fn layer(&self, depth: usize) -> Option<&[Block]> {
-        if depth < self.depth {
-            return Some(&self.buf[layer(depth)]);
-        } else if depth == self.depth {
-            return Some(&self.leaves);
+        match depth.cmp(&self.depth) {
+            Ordering::Less => Some(&self.buf[layer(depth)]),
+            Ordering::Equal => Some(self.leaves),
+            Ordering::Greater => None,
         }
-
-        None
     }
 
     /// Returns an iterator over the layers of the GGM tree.
--- a/crates/mpz-garble-core/src/lib.rs
+++ b/crates/mpz-garble-core/src/lib.rs
@@ -4,7 +4,6 @@
 
 #![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]
-#![allow(clippy::result_large_err)]
 
 pub(crate) mod circuit;
 mod evaluator;
--- a/crates/mpz-garble-core/src/lib.rs
+++ b/crates/mpz-garble-core/src/lib.rs
@@ -4,6 +4,7 @@
 
 #![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]
+#![allow(clippy::result_large_err)]
 
 pub(crate) mod circuit;
 mod evaluator;
--- a/crates/mpz-garble-core/src/store/evaluator.rs
+++ b/crates/mpz-garble-core/src/store/evaluator.rs
@@ -372,14 +372,20 @@ where
 /// Error for [`EvaluatorStore`].
 #[derive(Debug, thiserror::Error)]
 #[error("evaluator store error: {}", .0)]
-pub struct EvaluatorStoreError(#[from] ErrorRepr);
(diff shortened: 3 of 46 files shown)


crates/mpz-core/src/ggm.rs
```diff
@@ -139,13 +140,11 @@ impl<'a> GgmTree<'a> {
...
     /// Returns the layer at the given depth.
     pub fn layer(&self, depth: usize) -> Option<&[Block]> {
-        if depth < self.depth {
-            return Some(&self.buf[layer(depth)]);
-        } else if depth == self.depth {
-            return Some(&self.leaves);
+        match depth.cmp(&self.depth) {
+            Ordering::Less => Some(&self.buf[layer(depth)]),
+            Ordering::Equal => Some(self.leaves),
+            Ordering::Greater => None,
```
sinui0, line 146: what the heck :joy:

crates/mpz-garble-core/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 #![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]
+#![allow(clippy::result_large_err)]
```
sinui0: I would prefer to fix the error size than put in a crate wide allow

crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
```diff
@@ -219,6 +219,7 @@ pub(crate) mod state {
 
     pub(crate) trait State: sealed::Sealed {}
 
+    #[allow(clippy::large_enum_variant)]
```
sinui0: Better to box

crates/mpz-ot/src/chou_orlandi/receiver.rs
```diff
@@ -12,6 +12,7 @@ use serio::{stream::IoStreamExt as _, SinkExt as _};
 type Error = ReceiverError;
 
 #[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
```
sinui0: Box

crates/mpz-vm-core/src/call.rs
```diff
@@ -68,6 +68,7 @@ pub struct Call {
 
 impl Call {
     /// Creates a new call builder.
+    #[allow(clippy::new_ret_no_self)]
```
sinui0: rename to `builder`

sinui0 (changes requested):

crates/mpz-vm-core/src/call.rs
th4s: yeah, I decided against, because this is a public API, but ok will do. Will break tlsn I think.

crates/mpz-core/src/ggm.rs
th4s: Keep it or add a clippy allow?

crates/mpz-garble-core/src/store/evaluator.rs
```diff
@@ -372,14 +372,20 @@ where
 /// Error for [`EvaluatorStore`].
 #[derive(Debug, thiserror::Error)]
 #[error("evaluator store error: {}", .0)]
-pub struct EvaluatorStoreError(#[from] ErrorRepr);
+pub struct EvaluatorStoreError(#[source] Box<ErrorRepr>);
```
sinui0: Can we instead address why `ErrorRepr` is large?

crates/mpz-garble/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]
 #![forbid(unsafe_code)]
+#![allow(clippy::result_large_err)]
```
sinui0: this is no good

crates/mpz-ot-core/src/ferret/mpcot/sender.rs
```diff
@@ -196,6 +196,7 @@ pub(crate) mod state {
 
     pub(crate) trait State: sealed::Sealed {}
 
+    #[allow(clippy::large_enum_variant)]
```
sinui0: Box the hashes

sinui0 (changes requested):

crates/mpz-core/src/ggm.rs
sinui0: we can keep it i suppose

crates/mpz-garble-core/src/store/evaluator.rs
th4s: It is because `ErrorRepr::InconsistentFlush` contains `FlushView` twice. Each `FlushView` contains four `Vec`, which means 2 * 4 * 3 * 8 = 192 bytes. However I think the error variant is helpful because it tells you the view difference. If you do not want to box the error, we could box the `FlushView`s.

sinui0 (approved):

Issue #233, opened 2025-02-17
Add input adjustment bits to the transcript


themighty1:
Currently, the adjustment bits for the prover's input sent here 

https://github.com/privacy-scaling-explorations/mpz/blob/ba984a116e40cf669c597cfbe71beb71cdb729a0/crates/mpz-zk-core/src/lib.rs#L107

are not added to the transcript for Fiat-Shamir transform as they should be acc.to the quicksilver paper.

sinui0:
Closed by #241

PR #241, opened 2025-02-27
fix(zk): add missing adjustment bits to transcript for Fiat-Shamir.

yuroitaki:
https://github.com/privacy-scaling-explorations/mpz/issues/233

### DIFF
--- a/crates/mpz-zk-core/Cargo.toml
+++ b/crates/mpz-zk-core/Cargo.toml
@@ -24,6 +24,7 @@ blake3 = { workspace = true, features = ["serde"] }
 cfg-if = { workspace = true }
 rayon = { workspace = true, optional = true }
 itybity = { workspace = true }
+zerocopy = { workspace = true }
 
 [dev-dependencies]
 mpz-ot-core = { workspace = true, features = ["test-utils"] }
--- a/crates/mpz-zk-core/src/check.rs
+++ b/crates/mpz-zk-core/src/check.rs
@@ -30,7 +30,6 @@ pub(crate) struct Triple {
 
 #[derive(Debug, Default)]
 pub(crate) struct Check {
-    transcript: Hasher,
     triples: Vec<Triple>,
     adjust: BitVec<u8>,
 }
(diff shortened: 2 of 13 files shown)


crates/mpz-zk-core/src/check.rs
```diff
@@ -56,10 +55,10 @@ impl Check {
         !self.triples.is_empty()
     }
 
-    fn compute_chis(&self) -> Vec<Block> {
+    fn compute_chis(&self, transcript: &Hasher) -> Vec<Block> {
```
sinui0: Accept `chi` here instead

crates/mpz-zk-core/src/check.rs
```diff
@@ -91,9 +91,9 @@ impl Check {
             (u, v)
         }
 
-        self.transcript.update(self.adjust.as_raw_slice());
+        transcript.update(self.adjust.as_raw_slice());
```
sinui0, line 93: Can sample `chi` here then pass it into `compute_chis`

crates/mpz-zk-core/Cargo.toml
```diff
@@ -24,6 +24,7 @@ blake3 = { workspace = true, features = ["serde"] }
 cfg-if = { workspace = true }
 rayon = { workspace = true, optional = true }
 itybity = { workspace = true }
+zerocopy = { workspace = true }
```
sinui0, line 27: What's this for?

sinui0 (changes requested):
Looks great, but rather than store the hasher in the memory struct let's keep it outside and pass it into the memory flush methods.

yuroitaki:
> Looks great, but rather than store the hasher in the memory struct let's keep it outside and pass it into the memory flush methods.

If hasher is kept outside, doesn't that mean in `mpc-zk` crate, we need to modify some of the `Execute` trait [methods](https://github.com/privacy-scaling-explorations/mpz/blob/03249415727bd241634ad779376564a3dc55bc0b/crates/mpz-zk/src/prover.rs#L47) so that prover/verifier can access the same hasher?

crates/mpz-zk-core/Cargo.toml
yuroitaki: To convert input adjustment bits from `&[u32]` (`ProverFlush` storing it as `BitVec<u32>`) into `&[u8]` (expected by `hasher.update()`)
https://github.com/privacy-scaling-explorations/mpz/blob/a4476d88fe38bad210a89cf6366252f5ee28bc9b/crates/mpz-zk-core/src/store/prover.rs#L157

sinui0:
Yes, the hasher would be stored in the `Prover` and `Verifier` structs in `mpz-zk` and injected into the store flush and check methods

sinui0 (approved):
Good work!

PR #244, opened 2025-03-03
fix: cross-platform bitvec bug

sinui0:
This PR fixes a cross-platform issue arising from `bitvec`. The author of that crate recommends using only `BitVec<u32>` to mitigate this. Additionally, there is some variance in the backing store between platforms which busts transcript hashing. I've modified that code to only include the slice corresponding to initialized data.

### DIFF
--- a/crates/mpz-ot-core/src/ferret/spcot/receiver.rs
+++ b/crates/mpz-ot-core/src/ferret/spcot/receiver.rs
@@ -13,6 +13,7 @@ use mpz_core::{
     utils::{slices_from_lengths, slices_from_lengths_mut},
     Block,
 };
+use zerocopy::IntoBytes;
 
 use crate::{ferret::config::CSP, Derandomize};
 
@@ -89,7 +90,9 @@ impl SPCOTReceiver {
                 .map(|(b, m)| !b ^ m),
         );
 
-        self.transcript.update(flip.as_raw_slice());
+        let flip_len = flip.len();
+        self.transcript
+            .update(&flip.as_raw_slice().as_bytes()[..flip_len.div_ceil(8)]);
 
         Ok(Derandomize { flip })
     }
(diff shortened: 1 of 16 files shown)


crates/mpz-ot-core/src/ferret/spcot/receiver.rs
```diff
@@ -89,7 +90,9 @@ impl SPCOTReceiver {
                 .map(|(b, m)| !b ^ m),
         );
 
-        self.transcript.update(flip.as_raw_slice());
+        let flip_len = flip.len();
+        self.transcript
+            .update(&flip.as_raw_slice().as_bytes()[..flip_len.div_ceil(8)]);
```
th4s, line 95: I do not understand this line. Why `..flip_len.div_ceil(8)`

th4s (commented):
I have a question considering some adjustment.

crates/mpz-ot-core/src/ferret/spcot/receiver.rs
sinui0: There are 8 bits in a byte, this rounds up to ensure all bits are present

th4s: I was confused ^^, makes sense now

yuroitaki: > There are 8 bits in a byte, this rounds up to ensure all bits are present

is the cross platform issue here that, for different platforms, bitvec is internally storing the bits in words of different sizes, hence we have to always take the exact bytes here to ensure cross-platform consistency?

sinui0: yes

yuroitaki: > yes

cool~ from the PR description it also sounds like there is another cross platform issue besides the one I described above?

themighty1:
ack

Issue #248, opened 2025-03-04
ferret param selection error


themighty1:
Currently in https://github.com/privacy-scaling-explorations/mpz/blob/76e9e823a902fd615028635054c98e646450b00a/crates/mpz-ot-core/src/ferret/config.rs#L107

the corrected line should be `let net = param.n - cost;`

as reported in https://github.com/privacy-scaling-explorations/mpz/pull/202#discussion_r1933792895

yuroitaki:
Closing this as addressed by https://github.com/privacy-scaling-explorations/mpz/pull/249

PR #249, opened 2025-03-05
(fix) ferret lpn parameter selector

yuroitaki:
Discovered a bug when I ran tlsn's `attestation_prove` example in debug mode
```
thread '<unnamed>' panicked at /Users/yuroitaki/.cargo/git/checkouts/mpz-585819f1a9027d2e/68d0571/crates/mpz-ot-core/src/ferret/config.rs:107:19:
attempt to subtract with overflow
```
(P/S: in release mode there is no issue)

@sinui0 pointed out that this should be the fix — tested and all worked with the fix.


yuroitaki:
Found out that this also has been reported as per https://github.com/privacy-scaling-explorations/mpz/issues/248

th4s (approved):
:+1:

sinui0 (approved):

PR #252, opened 2025-03-08
2024 edition async closures

sinui0:
This PR updates `Context` to use async closures instead of scoped box futures.


th4s (commented):
:+1:

PR #253, opened 2025-03-08
refactor: update to rand 0.9

### LINKED ISSUE #232: Update `rand` to version `0.9`

sinui0:
This PR updates mpz to use rand 0.9. I've added a compatibility wrapper to support older versions of rand for a couple of our dependencies.

Closes #232


crates/mpz-core/src/rand.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Wraps `self` in a compatibility wrapper that implements `0.8` traits.
+    fn compat_by_ref(&mut self) -> Rand0_8CompatWrapper<&mut Self>
+    where
+        Self: Sized,
+    {
+        Rand0_8CompatWrapper(self)
+    }
+}
+
+impl<T> Rand0_8CompatExt for T where T: ?Sized {}
```
th4s, line 22: Shouldn't this have a bound on T, s.th like `where T: RngCore`?

th4s (commented):
Nice solution with compat extension trait. I agree that this approach is better than waiting for our dependencies to adapt.

crates/mpz-core/src/rand.rs
sinui0: #257

themighty1:
ack

PR #274, opened 2025-04-01
feat: parser accepts str

Meyanis95:
This PR refactors the circuit parser to accept str instead of requiring a file path, enabling the circuit parsing functionality to run in environments without a filesystem (like browsers or WebAssembly). It keeps the original `parse` method for backward compatibility, but implements it to use the new `parse_str` method internally.

### DIFF
--- a/crates/mpz-circuits/src/parse.rs
+++ b/crates/mpz-circuits/src/parse.rs
@@ -23,26 +23,22 @@ pub enum ParseError {
 }
 
 impl Circuit {
-    /// Parses a circuit in Bristol-fashion format from a file.
-    ///
-    /// See `https://nigelsmart.github.io/MPC-Circuits/` for more information.
+    /// Parses a circuit in Bristol-fashion format from a string.
     ///
     /// # Arguments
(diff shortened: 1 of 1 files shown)


crates/mpz-circuits/src/parse.rs
```diff
@@ -23,26 +23,22 @@ pub enum ParseError {
 }
 
 impl Circuit {
-    /// Parses a circuit in Bristol-fashion format from a file.
-    ///
-    /// See `https://nigelsmart.github.io/MPC-Circuits/` for more information.
```
th4s, line 28: This information was removed. I assume this was not on purpose?

sinui0:
lgtm modulo comment from @th4s , let's keep that info in there.

Could you please rebase onto the alpha.3 branch? There will be a small conflict

Meyanis95:
I was not able to resolve conflicts, I'm closing this PR and re-opening a clean one.
https://github.com/privacy-scaling-explorations/mpz/pull/275

Issue #239, opened 2025-02-25
Ferret SPCOT count inconsistent between platforms


sinui0:
When using Ferret in Uniform LPN mode I was getting errors related to SPCOT length mismatches between the Sender and Receiver. One party was running in WASM, other on x86. I suspect there is a difference in floating point arithmetic between the two builds or something of the like.

alupest:
@sinui0  Hi, I have send the PR https://github.com/privacy-scaling-explorations/mpz/pull/276 here, can you check?

heeckhau:
We investigated this to see if it is reproducible.

We wrote cross-platform regression tests with hardcoded reference values for `compute_table_length` (all production `t` values: 600, 1319, 2000) and `iteration_cost` (all 8 uniform LPN parameter sets). These are the only two places where float arithmetic (`(1.5 * (t as f32)).ceil()`) feeds into the SPCOT count path.

| Platform | `compute_table_length` | `iteration_cost` |
|---|---|---|
| ARM64 native (macOS) | pass | pass |
| WASM on ARM64 (wasmtime) | pass | pass |
| x86_64 native (Linux) | pass | pass |

All platforms produce identical results. The `f32` arithmetic is exact for these `t` values since both `t` and `1.5 * t` are exactly representable in `f32`.

**Conclusion:** We could not reproduce the original issue with current parameters on any tested platform combination. The bug may have been triggered by a previous set of parameters, a different WASM runtime (browser vs wasmtime), or a different code path that has since changed.

Closing as not reproducible.

PR #276, opened 2025-04-05
fix: SPCOT count inconsistency between platforms in Ferret Uniform LPN mode

alupest:
Replace platform-dependent `checked_next_power_of_two()` with a deterministic algorithm to ensure consistent SPCOT lengths across WASM and x86 platforms.

### DIFF
--- a/crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
+++ b/crates/mpz-ot-core/src/ferret/mpcot/receiver.rs
@@ -80,6 +80,8 @@ impl MPCOTReceiver<state::Initialized> {
                         let (_, pos) = buckets.get(x.value as usize)[x.hash_idx as usize];
                         idxs.push(pos);
                     } else {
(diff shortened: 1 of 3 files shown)


alupest:
@sinui0 This PR aims to resolve #239 
Let me know if the fix looks good

alupest:
@sinui0 Just a gentle reminder to review my PR and see if it resolves the issue tagged

sinui0 (commented):
I don't expect this to resolve the issue, as it would very surprising for unsigned integer ops to deviate across platforms. The issue is more likely from floating point arithmetic.

alupest:
@sinui0 I have tried to resolve the issue, please look into it. And if possible, please let me know how I might be able to reproduce the issue you faced so that I can address the issue better. Thanks

themighty1:
I'll close this PR. We should have a repro case first, then we could confirm if any given solution works.

Issue #234, opened 2025-02-20
Replace `bytemuck` with `zerocopy`


sinui0:

0x471:
Hi, can I work on this? @sinui0

sinui0:
> Hi, can I work on this? @sinui0 

That would be great. Please base on the alpha.1 branch for now

geooner:
hey @sinui0. Is this being working on ? and what's the status of this issue? Can I get the details and work on this?

mahmudsudo:
can i take on this ?

sinui0:
Anyone feel free to open a PR. The task is simple, replace all usage of `bytemuck` with `zerocopy`.

geooner:
Thank you @sinui0  for your response. Can I get some details about this before working on the issue? Do you have any suggestions?

adanimasuk:
@sinui0 I have sent a PR for this here #278

adanimasuk:
Hey @th4s , this is my PR about **replacing bytemuck with zerocopy**. Can you please take a look at my PR ? I’d really appreciate your feedback. Let me know if there’s anything you’d like me to adjust!

PR #278, opened 2025-04-14
Replace bytemuck with zerocopy

adanimasuk:
Replaced all usages of the `bytemuck` crate with `zerocopy`.

This change addresses Issue #234 by migrating transmutation logic to use `zerocopy`'s APIs for improved safety and maintenance. The `zerocopy` crate provides compile-time guarantees about the safety of transmutations where applicable.

### DIFF
--- a/crates/mpz-core/src/block.rs
+++ b/crates/mpz-core/src/block.rs
@@ -1,6 +1,5 @@
 //! A block of 128 bits and its operations.
 
-use bytemuck::{Pod, Zeroable};
 use clmul::Clmul;
 use core::ops::{BitAnd, BitAndAssign, BitXor, BitXorAssign};
 use generic_array::{typenum::consts::U16, GenericArray};
@@ -11,10 +10,11 @@ use std::{
     fmt::{Debug, Display},
     slice::from_raw_parts,
 };
+use zerocopy::{Immutable, IntoBytes, FromBytes};
 
 /// A block of 128 bits
 #[repr(transparent)]
-#[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize, Pod, Zeroable)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize, IntoBytes, FromBytes, Immutable)]
 pub struct Block([u8; 16]);
 
 impl Block {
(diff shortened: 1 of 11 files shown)


adanimasuk:
hey @sinui0, here is the PR for #234. Let me know if you have any questions. Please merge if everything looks right.
Thank You!

adanimasuk:
Hey @th4s, this is my PR about replacing `bytemuck` with `zerocopy`. Can you please take a look? I’d really appreciate your feedback. Let me know if there’s anything you’d like me to adjust!

crates/mpz-core/src/lpn.rs
th4s, line 1: Still some manual transmutes.

crates/mpz-core/src/prg.rs
```diff
@@ -179,7 +179,11 @@ impl Prg {
     /// Fill a block slice with random block values.
     #[inline(always)]
     pub fn random_blocks(&mut self, buf: &mut [Block]) {
-        let bytes: &mut [u8] = bytemuck::cast_slice_mut(buf);
+        let len = buf.len();
+        // SAFETY: Block is repr(transparent) over [u8; 16]
+        let bytes: &mut [u8] = unsafe {
+            std::slice::from_raw_parts_mut(buf.as_mut_ptr() as *mut u8, len * Block::LEN)
+        };
```
th4s, line 186: Manual transmute

crates/mpz-core/src/block.rs
```diff
@@ -142,13 +142,15 @@ impl Block {
...
     pub fn sigma(a: Self) -> Self {
-        let mut x: [u64; 2] = bytemuck::cast(a);
-        x[0] ^= x[1];
-        bytemuck::cast([x[1], x[0]])
+        let val = u128::from_le_bytes(a.0);
+        let low = (val & u64::MAX as u128) as u64;
+        let high = (val >> 64) as u64;
+        let new_low = low ^ high;
+        let result_val = ((new_low as u128) << 64) | (high as u128);
+        Block(result_val.to_le_bytes())
```
th4s, line 153: Is there an advantage in doing the shifting manually instead of using `zerocopy` like `bytemuck` has been used before here?

th4s (changes requested):
Thanks for the PR. I think this is a good start. 

Isn't the point of this PR to have `zerocopy` do the `unsafe` transmutation parts? I still see manual transmutations using `unsafe`. Is there a reason this is still done manually?

th4s:
Hmm, I now see, that what I ask for currently seems not supported for macros: https://github.com/google/zerocopy/issues/1817

But I think you should be able to use `FromBytes` for casting slices/DSTs

adanimasuk:
hey @th4s, Thanks for the feedback!

Regarding `sigma` function in `block.rs`.

The previous implementation used `bytemuck::cast` to reinterpret the `Block` as `[u64; 2]`. While `zerocopy` could certainly be used to perform a similar safe transmutation, the approach taken here was to switch to using standard `u128` arithmetic (via `from_le_bytes`, bitwise operations, and `to_le_bytes`).

The main advantage of this current manual `u128` approach is that it avoids type transmutation altogether for this specific operation. It relies only on standard library functions and safe bitwise arithmetic, making the extraction and recombination of the 64-bit parts explicit without needing `zerocopy`'s casting features here.

So, while `zerocopy` *could* replace `bytemuck::cast` for a similar transmutation-based implementation, the advantage of the current method is this avoidance of the transmutation step itself within the function's logic. The core XOR operation remains manual in either case.

PR #280, opened 2025-04-22
fix(co15): error when payload is too small

### LINKED ISSUE #194: bug(ot): co15 doesnt enforce correct ot count

themighty1:
This PR handles cases when co15 sender's payload is too small.

Closes #194


sinui0:
Please rebase onto `dev` as `alpha.3` has been merged into it

PR #285, opened 2025-04-25
fix(co15): error when payload is too small

### LINKED ISSUE #194: bug(ot): co15 doesnt enforce correct ot count

themighty1:
This PR handles cases when co15 sender's payload is too small.

Closes https://github.com/privacy-scaling-explorations/mpz/issues/194

### DIFF
--- a/crates/ot-core/src/chou_orlandi/receiver.rs
+++ b/crates/ot-core/src/chou_orlandi/receiver.rs
@@ -140,6 +140,11 @@ impl Receiver<state::Setup> {
                 payload.len(),
             ));
         }
+        // Check that all queued transfers can be processed.
+        let total_queued = self.queue.iter().map(|q| q.count).sum();
+        if total_queued > payload.len() {
+            return Err(ReceiverError::PayloadTooSmall(total_queued, payload.len()));
+        }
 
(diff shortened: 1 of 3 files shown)


crates/ot-core/src/chou_orlandi/receiver.rs
```diff
@@ -140,6 +140,11 @@ impl Receiver<state::Setup> {
                 payload.len(),
             ));
         }
+        // Check that all queued transfers can be processed.
+        let total_queued = self.queue.iter().map(|q| q.count).sum();
+        if total_queued > payload.len() {
+            return Err(ReceiverError::PayloadTooSmall(total_queued, payload.len()));
+        }
```
sinui0: Wouldn't it be equivalent to modify the above check to be `payload.len() != decryption_keys.len()`?

sinui0 (commented):
Adding a test case would be good

themighty1:
@sinui0 , ready for review

sinui0 (approved):

PR #286, opened 2025-04-25
make from_io in Context public

teddav:
While playing with `mpz` I needed to make 2 parties communicate over TCP and figured I needed to use Context::from_io which therefore needs to be public. See here: https://github.com/th4s/mpz-play/pull/1

### DIFF
--- a/crates/common/src/context.rs
+++ b/crates/common/src/context.rs
@@ -48,7 +48,12 @@ impl Context {
         }
     }
 
-    pub(crate) fn from_io(io: Io) -> Self {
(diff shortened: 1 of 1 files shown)


sinui0:
The public method is

https://github.com/privacy-scaling-explorations/mpz/blob/8f2efd63ba1d0c01b4e102dd46215a715e650f36/crates/common/src/context.rs#L35-L49

teddav:
wow... i'm blind 🤦‍♂️ thanks @sinui0 , im closing this

PR #287, opened 2025-04-25
feat: sha256 hash helper

sinui0:
This PR adds a helper for computing sha256 hashes.

I expect this to be removed in the future if/when we move to having an ISA.

### DIFF
--- a/crates/hash/src/sha256.rs
+++ b/crates/hash/src/sha256.rs
@@ -57,20 +57,39 @@ pub struct Sha256 {
 impl Sha256 {
     /// Creates a new SHA-256 hasher.
     pub fn new() -> Self {
-        Self {
-            state: None,
-            blocks: Vec::new(),
-            processed: 0,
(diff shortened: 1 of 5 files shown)


crates/hash/src/sha256.rs
```diff
@@ -0,0 +1,314 @@
...
+#[derive(Debug)]
+pub struct Sha256 {
+    state: Option<Array<U32, 8>>,
+    blocks: Vec<Block>,
+    processed: usize,
+}
+
+impl Sha256 {
+    /// Creates a new SHA-256 hasher.
+    pub fn new() -> Self {
```
th4s, line 59: This could be a `Default` implementation.

th4s (approved):
looking good :+1:

sinui0:
I updated it a bit which should help satisfy our requirements in tlsn

th4s (approved):

PR #290, opened 2025-04-30
feat(ot): adaptive shared RCOT

sinui0:
This PR updates the shared RCOT implementation to be adaptive so that instances can be dropped without deadlocking the remaining.

### DIFF
--- a/crates/common/src/sync/barrier.rs
+++ b/crates/common/src/sync/barrier.rs
@@ -16,6 +16,54 @@ pub struct AdaptiveBarrier {
     state: Arc<Mutex<State>>,
 }
 
+impl AdaptiveBarrier {
+    /// Creates a new barrier.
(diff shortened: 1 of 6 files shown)


crates/common/src/sync/barrier.rs
```diff
@@ -0,0 +1,357 @@
...
+            .sender
+            .send(BarrierResult::Leader(Waiters::new(waiters)))
+            .expect("leader should not be dropped before waking");
+
+        self.arrived = 0;
+        self.expected = 0;
+    }
+}
+
+impl AdaptiveBarrier {
```
th4s: Can we move this impl block below the struct definition?

th4s (approved):
nice!

themighty1 (commented):
ack

PR #294, opened 2025-05-13
feat(common): limit message size

th4s:
This PR adds a way to temporarily adjust the maximum frame size of `Io`.

### DIFF
--- a/crates/common/src/io.rs
+++ b/crates/common/src/io.rs
@@ -9,7 +9,10 @@ use bytes::{Bytes, BytesMut};
 use futures::{AsyncRead, AsyncWrite};
 use pin_project_lite::pin_project;
 use serio::{Framed, Sink, Stream, channel::MemoryDuplex, codec::Bincode};
-use tokio_util::{codec::LengthDelimitedCodec, compat::FuturesAsyncReadCompatExt as _};
+use tokio_util::{
+    codec::{Framed as TokioFramed, LengthDelimitedCodec},
+    compat::FuturesAsyncReadCompatExt as _,
+};
 
 trait Duplex:
     futures::Stream<Item = Result<BytesMut, std::io::Error>>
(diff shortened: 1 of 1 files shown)


crates/common/src/io.rs
```diff
@@ -23,6 +26,95 @@ impl<T> Duplex for T where
 {
 }
 
+trait FrameLimit {
```
sinui0: could just modify the `Duplex` trait instead of defining a new one. It's already private and only serves a purpose in this module.

crates/common/src/io.rs
```diff
@@ -44,19 +136,48 @@ impl Io {
...
+    /// * `frame_limit` - The new maximum frame length in bytes.
+    pub fn with_limit(&mut self, frame_limit: usize) -> WithLimit<'_> {
+        let old_limit = match &self.inner {
+            Inner::Transport { framed } => framed.inner().frame_limit(),
+            Inner::Memory { channel: _ } => None,
+        };
+
+        if let Inner::Transport { ref mut framed } = self.inner {
+            framed.inner_mut().set_frame_limit(frame_limit);
+        };
```
sinui0: These could be consolidated into one match block

crates/common/src/io.rs
```diff
@@ -23,6 +26,95 @@ impl<T> Duplex for T where
...
+trait FrameLimit {
+    /// Sets a new maximum frame length.
+    ///
+    /// # Arguments
+    ///
+    /// * `frame_limit` - The new maximum frame length in bytes.
+    fn set_frame_limit(&mut self, frame_limit: usize);
+
+    /// Returns the current frame limit, if available.
+    fn frame_limit(&self) -> Option<usize>;
```
sinui0: Why `Option`?

crates/common/src/io.rs
```diff
@@ -44,19 +136,48 @@ impl Io {
...
+    /// # Arguments
+    ///
+    /// * `frame_limit` - The new maximum frame length in bytes.
+    pub fn with_limit(&mut self, frame_limit: usize) -> WithLimit<'_> {
+        let old_limit = match &self.inner {
+            Inner::Transport { framed } => framed.inner().frame_limit(),
+            Inner::Memory { channel: _ } => None,
+        };
+
+        if let Inner::Transport { ref mut framed } = self.inner {
```
sinui0: imo using binding-mode operators like `ref mut` is uglier than just matching on `&mut self.inner`

crates/common/src/io.rs
```diff
@@ -23,6 +26,95 @@ impl<T> Duplex for T where
...
+
+    impl<'a> PinnedDrop for WithLimit<'a> {
+        fn drop(mut this: Pin<&mut Self>) {
+            let Some(old_limit) = this.old_limit else {
+                return;
+            };
+            let Inner::Transport { ref mut framed } = this.io.inner else {
+                return;
+            };
+            framed.inner_mut().set_frame_limit(old_limit);
```
sinui0:
```suggestion
            if let Inner::Transport { framed } = &mut this.io.inner {
                framed.inner_mut().set_frame_limit(self.old_limit.expect("old limit is always set for transport variant"));
            }
```

sinui0 (changes requested):
couple nits, otherwise lgtm

crates/common/src/io.rs
```diff
@@ -9,18 +9,99 @@ use bytes::{Bytes, BytesMut};
...
+        old_limit: Option<usize>,
+        #[pin]
+        io: &'a mut Io,
+    }
+
+    impl<'a> PinnedDrop for WithLimit<'a> {
+        fn drop(mut this: Pin<&mut Self>) {
+            let old_limit = this
+                .old_limit
+                .expect("old limit is always set for transport variant");
```
sinui0: :eyes:

sinui0 (changes requested):

crates/common/src/io.rs
th4s: good catch. the compiler disallows to borrow `this` because it is also borrowed as mutable when put inside the `if let`, so I put it before the `if let` which is obviously very wrong ^^

I propose to leave it as it initially was.

th4s: never mind I found a way

sinui0 (approved):

PR #217, opened 2025-02-04
add authenticated garbling

nakkstar123:
This PR adds the core authenticated garbling protocol from [WRK17](https://eprint.iacr.org/2017/030) in the trusted dealer model (assuming an ideal preprocessing functionality). The source code is added to mpz-garble-core/src/ in auth_gen.rs, auth_eval.rs and fpre.rs. Test code is in lib.rs.  Currently, fpre.rs is a substitute to generate the required preprocessing material and will be substituted with a secure implementation using COT in a future PR.

PR #298, opened 2025-05-16
Add authenticated garbling (rebased onto dev)

nakkstar123:
This PR adds optimized authenticated garbling (distributed half-gates from https://eprint.iacr.org/2018/578) to garble/ and garble-core/. The commits have been squashed and rebased onto dev as requested in #217.


sinui0:
Closing as per conversation about future scope of mpz and constrained resources

PR #299, opened 2025-05-19
fix: adjust frame limit for garbler and evaluator dependent on message size

### LINKED ISSUE #284: Calling `excute_all` causes `frame size too big`
This issues is linked to https://github.com/tlsnotary/tlsn/issues/801.

Our current VM implementation (semihonest garbling) runs into problems when what is sent over the wire during a flush is bigger than the default for `max_frame_length`, which is 8MB, of our `LengthDelimitedCodec`.

Possible solutions:
1) increase `max_frame_length` or
2) chunk VM flushes

th4s:
This PR sets the frame size for `GarblerFlush` and `EvaluatorFlush` depending on information from the synchronized `FlushView`.

Closes #284.

### DIFF
--- a/crates/garble-core/src/view.rs
+++ b/crates/garble-core/src/view.rs
@@ -1,8 +1,8 @@
 use mpz_memory_core::{
     Slice, View as ViewTrait, binary::Binary, correlated::Mac, view::VisibilityView,
 };
+use rangeset::{Difference, Disjoint, Intersection, Subset};
 use serde::{Deserialize, Serialize};
-use utils::range::{Difference, Disjoint, Intersection, Subset};
 
 type Range = std::ops::Range<usize>;
 type RangeSet = rangeset::RangeSet<usize>;
--- a/crates/garble-core/src/view.rs
+++ b/crates/garble-core/src/view.rs
@@ -1,6 +1,8 @@
-use mpz_memory_core::{Slice, View as ViewTrait, binary::Binary, view::VisibilityView};
-use rangeset::{Difference, Disjoint, Intersection, Subset};
+use mpz_memory_core::{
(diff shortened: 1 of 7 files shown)


themighty1 (approved):
ack

crates/garble/src/store/evaluator.rs
```diff
@@ -120,11 +120,21 @@ where
...
+            let evaluator_flush_size = flush.view().evaluator_flush_size();
+
             let (flush, ()) = ctx
                 .try_join(
-                    async |ctx| {
-                        ctx.io_mut().send(flush).await?;
-                        ctx.io_mut().expect_next().await.map_err(Error::from)
+                    async move |ctx| {
+                        ctx.io_mut()
+                            .with_limit(evaluator_flush_size)
```
sinui0, line 130: There is no need to adjust the limit while sending, it is only used for deserialization.

crates/garble-core/src/view.rs
```diff
@@ -50,6 +53,28 @@ pub(crate) struct FlushView {
 }
 
 impl FlushView {
+    /// Returns the expected garbler flush size in bytes.
+    pub fn garbler_flush_size(&self) -> usize {
+        let macs = self.macs.len() * Mac::BIT_SIZE;
+        let decode_info = self.decode_info.len();
```
sinui0, line 59: This will estimate low as there is packing overhead in `BitVec`

themighty1:
tagging @th4s to take a look at those comments

crates/garble/src/store/evaluator.rs
th4s: Both `Encoder` and `Decoder` check this: https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/codec/length_delimited.rs#L608

PR #302, opened 2025-06-05
new banner

heeckhau:


sinui0:
I like the existing banner. This new image is more of a logo

PR #305, opened 2025-06-06
fix(garble): bump estimated flush size

themighty1:
This PR bumps the flush size estimation cushion to 1024 bytes.

After https://github.com/privacy-scaling-explorations/mpz/pull/299 was merged, I'm already seeing `frame size too big` errors 
coming from https://github.com/tokio-rs/tokio/blob/2440d113ff9841060b1876628c1153f0bcf2945c/tokio-util/src/codec/length_delimited.rs#L1082

The error went away when I bumped the cushion.

While we try to figure out the best estimation algorithm, let's just temporarily bump the value.


sinui0:
We should just address this issue correctly now

sinui0 (approved):

PR #306, opened 2025-06-07
refactor(garble): remove view from flush message

sinui0:
This PR removes the view from the flush messages in the garble vm, as it is unnecessary. I've also adjusted the message limit logic to account for this.

### DIFF
--- a/crates/garble-core/src/view.rs
+++ b/crates/garble-core/src/view.rs
@@ -55,16 +55,17 @@ pub struct FlushView {
 impl FlushView {
     /// Returns the expected garbler flush size in bytes.
     pub fn garbler_flush_size(&self) -> usize {
-        let macs = self.macs.len() * Mac::BIT_SIZE;
-        let decode_info = self.decode_info.len();
-        let mac_commitments = self.decode_info.len() * Mac::BIT_SIZE * 2;
+        let macs = self.macs.len() * Mac::BYTE_SIZE;
+        let decode_info = self.decode_info.len().div_ceil(8);
+        let mac_commitments = self.decode_info.len() * Mac::BYTE_SIZE * 2;
 
         // Add 1024 bytes of cushion room.
-        (macs + decode_info + mac_commitments) / 8 + self.size() + 1024
+        macs + decode_info + mac_commitments + 1024
     }
 
     /// Returns the expected evaluator flush size in bytes.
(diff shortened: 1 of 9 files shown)


crates/garble-core/src/view.rs
```diff
@@ -55,24 +55,25 @@ pub struct FlushView {
...
+        let mac_commitments = self.decode_info.len() * Mac::BYTE_SIZE * 2;
 
         // Add 1024 bytes of cushion room.
-        (macs + decode_info + mac_commitments) / 8 + self.size() + 1024
+        macs + decode_info + mac_commitments + 1024
     }
 
     /// Returns the expected evaluator flush size in bytes.
     pub fn evaluator_flush_size(&self) -> usize {
+        let decode_info = self.decode_info.len().div_ceil(8);
```
th4s, line 68: This should be `self.decode` as before?

crates/garble/src/store/evaluator.rs
```diff
@@ -126,18 +126,16 @@ where
...
             let mut cot = self.core.acquire_cot();
 
-            let garbler_flush_size = flush.view().garbler_flush_size();
-            let evaluator_flush_size = flush.view().evaluator_flush_size();
-
+            let expected_size = self.core.flush_view().garbler_flush_size();
             let (flush, ()) = ctx
                 .try_join(
                     async move |ctx| {
+                        ctx.io_mut().send(flush).await?;
```
th4s, line 133: `Encoder` also checks for overflow, so we should also use the adjusted limit for sending.
https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/codec/length_delimited.rs#L606-L613

crates/garble/src/store/garbler.rs
```diff
@@ -126,18 +126,16 @@ where
...
             let mut cot = self.core.acquire_cot();
 
-            let garbler_flush_size = flush.view().garbler_flush_size();
-            let evaluator_flush_size = flush.view().evaluator_flush_size();
-
+            let expected_size = self.core.flush_view().evaluator_flush_size();
             let (flush, ()) = ctx
                 .try_join(
                     async move |ctx| {
+                        ctx.io_mut().send(flush).await?;
```
th4s, line 133: same

th4s (commented):
I think here are some things off which deserve a fix/discussion.

Issue #176, opened 2024-08-07
implement a fix for KOS15


themighty1:
Our KOS15 implementation is based on the an old KOS15 paper whose proof was recently proven invalid.
The KOS15 paper was updated [https://eprint.iacr.org/2015/546](https://eprint.iacr.org/2015/546.pdf) and now contains a fix in Figure 10.

The fix seems straightforward to implement. The paper says 
"We have not included various optimizations for reducing communication and computation.". But even in its
unoptimized state, the Protocol in Figure 10 will be performant enough for bootstrapping Ferret OT.

sinui0:
This will reduce a round of communication too, removing the cointoss protocol.

robinhundt:
One thing to note is that the new consistency check in the updated paper is best performed on the untransposed Q and T matrices.

I've recently implemented the new consistency check in my own library (https://github.com/robinhundt/CryProt/pull/6 if anyone's curious) and also optimized the performance a little bit. Initially, I was traversing the columns of T and Q, multiplying by the same chi per iteration. However, I got better performance by iterating over the rows of T and Q and therefore also iterating 128 times over all values of chi_i. I'm fairly certain that the increase in performance is due to less cache-misses due to reading T and Q sequentially. Maybe this is of help to you as well :)

PR #315, opened 2025-07-29
fix(ot): implement a fixed KOS check

themighty1:
This PR implements a fixed KOS check discussed in https://github.com/privacy-scaling-explorations/mpz/issues/176

### DIFF
--- a/crates/ot-core/src/kos/receiver.rs
+++ b/crates/ot-core/src/kos/receiver.rs
@@ -6,7 +6,7 @@ use crate::{
     rcot::{RCOTReceiver, RCOTReceiverOutput},
 };
 
-use itybity::{FromBitIterator, IntoBits};
+use itybity::{BitLength, FromBitIterator, IntoBitIterator, IntoBits};
 use mpz_common::future::{MaybeDone, Sender, new_output};
 use mpz_core::{Block, prg::Prg};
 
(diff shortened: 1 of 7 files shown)


crates/ot-core/src/kos/receiver.rs
```diff
@@ -190,50 +199,62 @@ impl Receiver<state::Extension> {
...
+        // Sample random weights.
+        let chis: Vec<Block> = (0..m).map(|_| rng.random()).collect::<Vec<_>>();
+
+        fn compute_rlc(blocks: &[Block], chis: &[Block]) -> Block {
+            let (a, b) = blocks
+                .iter()
+                .zip(chis)
+                .map(|(q, chi)| q.clmul(*chi))
+                .reduce(|(_a, _b), (a, b)| (a ^ _a, b ^ _b))
+                .expect("iterator is not empty");
```
th4s: There is a `Block::gfmul` method. We could use it to simplify here.

crates/ot-core/src/kos/receiver.rs
```diff
@@ -386,17 +410,23 @@ pub mod state {
...
 
-        /// Receiver's unchecked ts
-        pub(super) unchecked_ts: Vec<Block>,
-        /// Receiver's unchecked choices
+        /// Receiver's unchecked ts before transposing.
+        ///
+        /// This is a row-major matrix which has [CSP] rows.
+        pub(super) unchecked_ts: Vec<Vec<Block>>,
+
+        /// Receiver's unchecked qs after transposing.
```
th4s:
```suggestion
        /// Receiver's unchecked ts after transposing.
```

th4s (approved):
:+1: nice, lgtm

crates/ot-core/src/kos/receiver.rs
themighty1: thanks, i realized that we have `Block::inn_prdt_red` which is even better

PR #317, opened 2025-08-07
fix(ot): improve ideal functionality performance

themighty1:
This PR fixes the performance of the ideal functionality which significantly affected the benchmarks:

- uses separate mutexes for each field, so that we don't have a situation where the sender and the receiver wait on each other to release the lock
- uses split_off instead of drain where possible

The performance problem is more obvious if you set in mpz/crates/zk/benches/zk.rs

```const BLOCK_COUNT: usize = 1000;```

and run
```
cargo bench --features="rayon" --bench=zk -- --sample-size=10
```

Before this PR:
zk/aes128               time:   [8.9342 s 9.1142 s 9.3012 s]

After this PR:
zk/aes128               time:   [1.4300 s 1.4394 s 1.4493 s]

### DIFF
--- a/crates/ot-core/src/ideal/rcot.rs
+++ b/crates/ot-core/src/ideal/rcot.rs
@@ -35,20 +35,20 @@ struct ReceiverState {
 /// Ideal RCOT functionality.
 #[derive(Debug, Clone)]
 pub struct IdealRCOT {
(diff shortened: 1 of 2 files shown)


th4s (approved):
:+1:

PR #318, opened 2025-08-07
feat(ot): add QuickSilver batch gate check

### LINKED ISSUE #200: feat: QS consistency check buffer limit
So that means if the circuit is huge, we have to handle very long vectors of F_2^{128} for the check, which may cause memory issues?

_Originally posted by @xiangxiecrypto in https://github.com/privacy-scaling-explorations/mpz/issues/198#issuecomment-2574266028_

Let's add a limit to the check buffer which causes the parties to execute the consistency check before using too much memory.

themighty1:
This PR adds a feature to set the AND gate buffer for the consistency check.

### DIFF
--- a/crates/zk/src/callstack.rs
+++ b/crates/zk/src/callstack.rs
@@ -1,5 +1,5 @@
 use mpz_vm_core::{Call, memory::Slice};
-use std::{ops::RangeBounds, slice::Iter, vec::IntoIter};
+use std::slice::Iter;
 
 #[derive(Debug, Default)]
 pub(crate) struct CallStack {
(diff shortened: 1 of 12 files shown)


sinui0:
Can you please rebase? Also, the `config` module was not checked-in

themighty1:
@sinu, done

crates/zk/src/callstack.rs
```diff
@@ -0,0 +1,41 @@
...
+        F: FnMut(&mut (Call, Slice)) -> bool,
+        R: RangeBounds<usize>,
+    {
+        let extracted = self.inner.extract_if(range, filter).collect::<Vec<_>>();
+
+        let and_gates: usize = extracted.iter().map(|(c, _)| c.circ().and_count()).sum();
+        self.and_count -= and_gates;
+
+        extracted.into_iter()
+    }
```
sinui0: We can encapsulate some more logic here and make it simpler.

```rust
/// Returns the ready calls based on the provided predicate.
///
/// # Arguments
///
/// * `soft_limit` - Calls will stop being yielded after this limit of AND gates is met or exceeded.
/// * `f` - Predicate determining whether a call is ready.
pub(crate) fn extract_if<F: Fn(&Call) -> bool>(&mut self, soft_limit: usize, f: F) -> impl Iterator<Item = (Call, Output)> {
    let mut total = 0;
    let mut limit = false;
    self.inner.extract_if(.., |(call, _)| {
        if !limit && f(call) {
            let and_count = call.circ().and_count();
            self.and_count -= and_count;
            total += and_count;
            limit = total >= soft_limit;

            true
        } else {
            false
        }
    })
}
```

crates/zk/src/prover.rs
```diff
@@ -106,13 +110,25 @@ where
 
     async fn execute(&mut self, ctx: &mut Context) -> VmResult<()> {
         let mut prover = Core::default();
+
+        // How many gates are to be checked.
+        let mut to_be_checked = 0;
```
sinui0: This should be exposed by the prover and verifier core structs.

Eg.

```rust
impl Prover {
  /// Returns the number of pending gates that need to be checked.
  pub fn pending(&self) -> usize {}
}
```

crates/zk/src/prover.rs
```diff
@@ -182,9 +198,29 @@ where
                     .set_output_macs(output, &output_macs)
                     .map_err(VmError::memory)?;
             }
+
+            if to_be_checked >= self.config.batch_size() {
```
sinui0:
```suggestion
            if prover.pending() >= self.config.batch_size() {
```

sinui0 (approved):
Looks good, added a couple suggestions which should help with clarity

themighty1:
thanks, all feedback added, merging

PR #319, opened 2025-08-18
Ferret: consider available OTs when allocating more than once

th4s:
Currently Ferret [always extends](https://github.com/privacy-scaling-explorations/mpz/blob/dev/crates/ot-core/src/ferret/sender.rs#L84-L87) when new OTs are requested via `alloc`, although it could satisfy the requested amount with left-over OTs from previous extensions.

This PR changes the Ferret implementation so that available OTs are used when allocating again after an iteration. I noticed when benching TLSNotary with our harness that Ferret extended 3x although 2x is more than sufficient.

It also changes that from now on available OTs are considered for bootstrapping.

### DIFF
--- a/crates/ot-core/src/ferret/receiver.rs
+++ b/crates/ot-core/src/ferret/receiver.rs
@@ -35,6 +35,7 @@ struct Queued {
 pub struct Receiver<COT> {
     cot: Arc<Mutex<COT>>,
     alloc: usize,
+    missing: usize,
     queue: VecDeque<Queued>,
     transfer_id: TransferId,
     prg: Prg,
@@ -54,6 +55,7 @@ where
         Self {
             cot: Arc::new(Mutex::new(cot)),
             alloc: 0,
+            missing: 0,
             queue: VecDeque::new(),
             transfer_id: TransferId::default(),
             prg: Prg::from_seed(seed),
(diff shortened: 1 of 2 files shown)


crates/ot-core/src/ferret/receiver.rs
```diff
@@ -102,11 +102,11 @@ where
 
     /// Allocates COTs for bootstrapping.
     pub fn alloc_bootstrap(&self) -> Result<()> {
-        let cost = self.config.bootstrap_cost();
+        let missing = self.config.bootstrap_cost() - self.macs.len();
```
themighty1: let's use `saturating_sub` here

crates/ot-core/src/ferret/receiver.rs
```diff
@@ -119,7 +119,8 @@ where
         };
 
         // If COTs are empty, we haven't bootstrapped from inner COT yet.
```
themighty1: i think this comment needs adjusting, it was originally incorrect, we can also get here when `self.config.reserve_bootstrap() is false`

crates/ot-core/src/ferret/sender.rs
```diff
@@ -103,11 +103,11 @@ where
 
     /// Allocates COTs for bootstrapping.
     pub fn alloc_bootstrap(&self) -> Result<()> {
-        let cost = self.config.bootstrap_cost();
+        let missing = self.config.bootstrap_cost() - self.keys.len();
```
themighty1: let's use saturating_sub here

crates/ot-core/src/ferret/sender.rs
```diff
@@ -120,25 +120,25 @@ where
         };
 
         // If COTs are empty, we haven't bootstrapped from inner COT yet.
```
themighty1: need to adjust the comment

themighty1 (approved):
good catch, lgtm, modulo some minor things

PR #259, opened 2025-03-15
refactor(circuits): delete macros and simplify

sinui0:
This PR removes typing information from `mpz-circuits` and deletes the macro crate.

PR #320, opened 2025-08-19
docs(readme): remove broken link to circuits-macros

Galoretka:
Don't know if you accept changes like this, coudn't find infromation about in in contributions.md
but the link is indeed broken, i couldn't find a way to replace it with the correct, so decided to remove it


heeckhau:
This was deleted in https://github.com/privacy-scaling-explorations/mpz/pull/259

heeckhau (approved):

PR #321, opened 2025-08-20
Ferret: correctly estimate params when not sufficient OTs available

th4s:
This PR fixes that the current `default_param_selector` does not correctly return valid parameters when the amount of available OTs is too small. The `if available <= cost` branch returns the `params` of the current iteration, but it should return those of the iteration before.

### DIFF
--- a/crates/ot-core/src/ferret/config.rs
+++ b/crates/ot-core/src/ferret/config.rs
@@ -107,14 +107,14 @@ fn default_parameter_selector(ty: LpnType, available: usize, additional: usize) 
         let cost = iteration_cost(ty, *param);
         let net = param.n - cost;
 
-        // Only select params for which we have enough OTs available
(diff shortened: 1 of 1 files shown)


crates/ot-core/src/ferret/config.rs
```diff
@@ -102,12 +102,20 @@ fn default_parameter_selector(ty: LpnType, available: usize, additional: usize)
...
-        if available <= cost || net >= additional {
+
+        // Only selects params for which we have enough OTs available.
+        if available < cost {
+            return last_valid_param;
+        } else {
+            last_valid_param = *param;
+        }
+
+        // Returns the smallest params that satisfy the additionaly requested amount.
```
themighty1:
```suggestion
        // Returns the smallest params that satisfy the additionally requested amount.
```

themighty1 (approved):
good find!

PR #328, opened 2025-09-23
feat(hash): add blake3 hash

yuroitaki:
https://github.com/tlsnotary/tlsn/issues/882


themighty1 (approved):
awesome work, just need to fix clippy lints and it's ok to merge

PR #334, opened 2025-10-07
Convert otp.rs test to compile-time API check

VolodymyrBg:
Replace a vacuous #[test] in crates/vm-core/src/otp.rs with a compile-time API conformance check. The previous test defined inner functions that were never called and contained no assertions, producing a misleading green test without runtime verification. This change preserves the intended compile-time validation of OneTimePad::mask_private for U8 and Vector<U8> while avoiding false test coverage and aligning with repository testing practices.

### DIFF
--- a/crates/vm-core/src/otp.rs
+++ b/crates/vm-core/src/otp.rs
@@ -78,9 +78,10 @@ mod tests {
 
     use super::*;
 
-    #[test]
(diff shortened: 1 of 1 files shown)


th4s:
Thanks, but decided not to merge since this is no improvement as it actually makes it worse, by producing a lint warning.

PR #335, opened 2025-10-08
feat: add ideal-vm crate

themighty1:
This PR adds an ideal VM implementation to be used for testing.


sinui0 (approved):
Excellent, thank you!

PR #339, opened 2025-10-09
feat(garble): randomize gate id

themighty1:
This PR implements gate randomization for garbling.

The rationale for it is discussed  https://github.com/privacy-ethereum/mpz/issues/7

### DIFF
--- a/crates/garble-core/src/lib.rs
+++ b/crates/garble-core/src/lib.rs
@@ -20,8 +20,13 @@ pub use garbler::{
     EncryptedGateBatchIter, EncryptedGateIter, Garbler, GarblerError, GarblerOutput,
 };
 pub use mpz_memory_core::correlated::{Delta, Key, Mac};
-use serde::{Deserialize, Serialize};
 pub use view::FlushView;
+
+use rand::{
+    Rng,
+    distr::{Distribution, StandardUniform},
+};
+use serde::{Deserialize, Serialize};
 
 const KB: usize = 1024;
 const BYTES_PER_GATE: usize = 32;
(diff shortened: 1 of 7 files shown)


crates/garble-core/src/lib.rs
```diff
@@ -37,6 +43,22 @@ const MAX_BATCH_SIZE: usize = 4 * KB;
 /// This puts an upper limit on that waste.
 pub(crate) const DEFAULT_BATCH_SIZE: usize = MAX_BATCH_SIZE / BYTES_PER_GATE;
 
+/// The initial gate id.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
th4s, line 47: This can be `Copy`, then we do not need to clone it.
```suggestion
#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
```

th4s (approved):
lgtm :+1:

sinui0:
I don't think I'm a fan of this change. Reading that paper the issue is the re-use of a tweak (`gid`). So the task is to ensure that a tweak is never reused. Rather than picking a starting `gid` for each circuit, I would prefer to maintain a global counter.

themighty1:
@sinui0, can you elaborate how a global counter would work?
( keeping in mind that `gid` must be unique even across multiple independent sessions)

sinui0:
> keeping in mind that `gid` must be unique even across multiple independent sessions

This is not clear to me, I'll need a definition of "session". The paper highlights an attack for recovering `R` (i.e. `Δ`) when the tweaks are reused. We never use the same `Δ` with multiple garbler instances.

It seems to me it is sufficient to maintain the counter per _VM instance_ (what I meant by global) not per circuit. The circuit boundary is unnecessary, we just need to maintain a counter for the total number of AND gates garbled, which can start at 1 and be maintained implicitly by both parties.

If we really want to keep the randomization to mitigate the attack when a `Δ` is accidentally reused with multiple VMs, we could randomize the initial global counter.

themighty1:
I see, you are suggesting randomizing the initial gate id and then incrementing per __VM instance__. That should work.

(the attack works even if different deltas are used across multiple circuits https://github.com/privacy-ethereum/mpz/issues/7#issuecomment-2538444039 )

themighty1:
Closed in favour of https://github.com/privacy-ethereum/mpz/pull/348

PR #347, opened 2025-10-31
feat(circuits): added AES pre/post key schedule circuits

themighty1:
This PR adds circuits to compute AES key schedule and AES encryption with key schedule.

Additionally:

- pre-build binary circuits are now provided by the `mpz-circuits-data` crate. 
Before this, the binary circuits were built manually, now via a build script.

- tidies up cargo configs

- removes stray chars from bristol/aes_128_reverse.txt


crates/circuits/circuits/bristol/aes_128_reverse.txt
```diff
@@ -8180,7 +8180,7 @@
 2 1 1519 11996 12126 XOR
 2 1 12121 12125 12114 AND
 2 1 12114 12037 12040 XOR
-2 1 122 120 12113 AND
+2 1 12122 12120 12113 AND
```
sinui0, line 8183: What happened here?

crates/circuits/src/parse.rs
```diff
@@ -219,6 +221,33 @@ mod tests {
...
+        let circ = Circuit::parse("circuits/bristol/aes_128_post_key_schedule.txt").unwrap();
+        let (_, ks, msg, expected_out) = aes_vectors();
+
+        let out = <[u8; 16]>::from_lsb0_iter(
+            circ.evaluate(ks.iter_lsb0().chain(msg.iter_lsb0()))
+                .unwrap(),
+        );
+
+        assert_eq!(expected_out, out);
+    }
```
sinui0, line 249: These don't need to be in this module. The existing tests are for testing the _parsing_ not the circuits themselves.

sinui0 (changes requested):
I support addressing the build issue but it should be in a separate PR.

For the build issue, I would suggest renaming the existing `mpz-circuits` crate to `mpz-circuits-core`. I would add a new third crate called `mpz-circuits` which depends on `mpz-circuits-core` and `mpz-circuits-data` which only re-exports the core modules as well as statically including the pre-built circuits from the data crate.

crates/circuits/circuits/bristol/aes_128_reverse.txt
themighty1: i'll open a new PR just  to fix the AES circuit and explain there

themighty1:
There were too many conflicts to rebase this PR on dev, so I'm closing this. All feedback was addressed in separate PRs.

PR #348, opened 2025-11-03
feat(garble-core): randomize gate id and cipher key

themighty1:
This PR adds randomization to garbler's initial gate id and key.

The rationale for it was discussed in:
https://github.com/privacy-ethereum/mpz/issues/341
https://github.com/privacy-ethereum/mpz/issues/7


crates/garble-core/src/evaluator.rs
```diff
@@ -79,6 +126,86 @@ impl Evaluator {
...
+impl State {
+    pub(crate) fn take(&mut self) -> State {
+        std::mem::replace(self, State::Error)
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Setup {
+    /// The id to be assigned to the next evaluated AND gate.
+    pub current_gid: u128,
```
sinui0: Fields should almost never be `pub`

```suggestion
    current_gid: u128,
```

themighty1:
@sinui0 , could you ack the conceptual approach of this PR? Then I can resolve the conflicts.

crates/garble-core/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio = { workspace = true, features = ["sync"] }
 [dev-dependencies]
 mpz-ot-core = { workspace = true, features = ["test-utils"] }
 
+bincode = { workspace = true }
```
sinui0:
```suggestion
```

crates/garble-core/src/garbler.rs
```diff
@@ -144,6 +305,7 @@ pub struct EncryptedGateIter<'a, I> {
     outputs: Range<usize>,
     /// Whether the entire circuit has been garbled.
     complete: bool,
+    pd: PhantomData<&'a ()>,
```
sinui0, line 311: Is this necessary?

crates/garble/src/protocol/semihonest/evaluator.rs
```diff
@@ -301,21 +326,43 @@ where
             self.execute_preprocessed()?;
         }
 
+        {
+            let mut core = self.core.try_lock().unwrap();
+            if !core.is_setup() {
+                let msg: SetupMsg = ctx.io_mut().expect_next().await?;
+                core.setup(msg).expect("core was not already set up");
```
sinui0: Safer to return an error in case the logic behind this assumption changes, e.g. validation in the setup message

crates/garble/src/store/garbler.rs
```diff
@@ -25,10 +25,6 @@ impl<COT> GarblerStore<COT> {
         }
     }
 
-    pub(crate) fn delta(&self) -> &Delta {
```
sinui0, line 28: Is this no longer used? I feel it should persist

sinui0 (approved):

crates/garble-core/src/garbler.rs
themighty1: yes, the iterator type I requires it (I: Iterator<Item = &'a Gate>)

crates/garble/src/store/garbler.rs
themighty1: The architecture changed fundamentally:

  1. Before: Each time you garble, you'd retrieve delta from the store and pass it to a new GarblerCore instance
  2. After: A single persistent Core is created with the delta, and it allocates workers that already have access to the delta

crates/garble-core/Cargo.toml
themighty1: we use bincode in a test which serializes and deserializes the SetupMsg to simulate a realistic scenario where the setup message is transmitted

sinui0: That is essentially just testing serde, I would omit that in unit tests unless you're specifically testing a custom serialization impl

themighty1:
Addressed feedback, merged dev.
I realized that key/gid must be derived from a seed, so that required API change (one of the last commits)
ready for review @sinui0

crates/garble-core/Cargo.toml
themighty1: Sorry, I misspoke, it is actually used in crates/garble-core/benches/garble.rs
```
b.iter(|| {
            let setup: SetupMsg = bincode::deserialize(&msg).unwrap();
            let mut ev = Evaluator::default();
            ev.setup(setup).unwrap();
```
I used it to avoid making `SetupMsg` Clone/Copy.

(oops, I noticed that I made `SetupMsg` Clone and Copy already which I shouldn't have)

do you agree with keeping bincode and removing Clone/Copy ?

sinui0 (approved):

crates/garble-core/Cargo.toml
themighty1: @sinui0 , i think you may have missed my last question here.

sinui0: I would just keep it `Clone`. Nothing wrong with that. `Copy` is too presumptuous. The serialization is unnecessary, IMO, but its just a nit

PR #357, opened 2025-12-11
feat: add predicate crate

themighty1:
This PR adds the `predicate` crate as well some useful circuit ops.

The predicate crate contains:
- compiler to a boolean circuit
- pre-built predicates for HTTP and JSON validation

This is enough to get the ball rolling. We will need to add more pre-built predicates later and maybe move them under the `tlsn` crate.


sinui0:
Closing in favor of WASM VM path

PR #358, opened 2025-12-12
feat(memory-core): add Bool type

themighty1:
This PR adds a VM memory bool type which is useful when checking that a predicate evaluates to true or false.


sinui0:
I'm wary of adding this because memory is currently supposed to be byte indexed, even though it is actually bit indexable. These bindings will go away if we move to WASM, so let's pause for now.

themighty1:
agreed

PR #359, opened 2025-12-19
feat: add wasm benches

themighty1:
This PR introduces a WASM harness for benchmarking mpz primitives and refactors existing benchmarks to measure each primitive in isolation from its counterparty

#### Refactored ideal OT functionalities to message-passing pattern                                          
  - Converted ideal OT/COT/ROT/RCOT to use message-passing instead of direct state sharing                
  - Benchmarked against previous approach - no performance degradation observed        

 #### Added context recording/replaying infrastructure to test-utils                                          
  - Benchmarked against current test contexts to confirm no overhead                                      

 #### CI updates                                                                                              
  - Excludes mpz-wasm-bench from native clippy (requires wasm32 target)


.github/workflows/rust.yml
```diff
@@ -58,7 +58,14 @@ jobs:
         run: cargo +nightly fmt --check --all
 
       - name: Run Clippy
-        run: cargo +nightly clippy --all-targets --all-features -- -D warnings
+        run: cargo +nightly clippy --workspace --exclude mpz-wasm-bench --all-targets --all-features -- -D warnings
+
+      - name: Install wasm32 target
```
heeckhau: You can avoid this step by adding `          targets: wasm32-unknown-unknown
` to `dtolnay/rust-toolchain@stable`

heeckhau (approved):
ack

PR #379, opened 2026-04-17
feat(mpz-fields): add GF(2^64) field

themighty1:
This PR implements the GF(2^64) field

This field will be used in the upcoming Quicksilver polynomial proof protocol.


crates/fields/src/test_data.rs
sinui0, line 1: This seems overly complicated and adds noise

sinui0 (changes requested):
Testing against swanky seems overkill, can we just have a set of test vectors?

themighty1:
ref values added, ready for review

sinui0 (approved):

PR #381, opened 2026-04-17
fix(wasm-bench): enable web-sys Worker feature for web-spawn

themighty1:
web-spawn 0.2.1 uses web_sys::Worker but relies on downstream crates
  to enable the "Worker" cargo feature of web-sys. Adding it (along with
  WorkerOptions and WorkerType, which Worker::new_with_options needs)
  restores the wasm32 clippy build which is currently failing.


heeckhau (approved):
lgtm

PR #382, opened 2026-04-23
feat(fields): constant time, tests, accelerated inner product

sinui0:
## Summary

  Finite-field arithmetic overhaul of `crates/fields`. Adds per-architecture accelerated backends for
  Gf2_64 and Gf2_128, a trait-level `inner_product` + `square` with SIMD-accelerated per-field overrides,
  an optional `rayon` feature that auto-parallelises large inner products, and substantially expanded test
  and benchmark coverage.

  ## What's new

  - **PCLMULQDQ backend for Gf2_128** — mirrors the existing Gf2_64 PCLMULQDQ path. 4 CLMULs + 2-round
  carry-less reduction; replaces the per-bit schoolbook `while (x != 0) && (y != 0)` loop.
  - **WASM `simd128` backend** for both fields. No carry-less-multiply in stable wasm-simd, so uses a
  lane-parallel BearSSL bit-interleaving trick (forward + bit-reversed halves in v128 lanes via
  `i64x2_mul`).
  - **Portable soft backend** for targets without PCLMUL or WASM SIMD. Uses BearSSL `bmul64_full` — a
  constant-time 64×64 → 128-bit carry-less multiply that works on any platform with constant-time scalar
  `mul`. Replaces the previous variable-time schoolbook loop.
  - **Dispatch via `cfg_select!`**: each field has `x86`, `wasm`, `soft` submodules, picked at compile time
   in one block. Same pattern used for the `rayon` feature branch.
  - **`Field::inner_product`** — new trait method with default fold + per-field SIMD override
  (`inner_product_chunk`). Each backend's implementation accumulates carry-less partials without reducing
  and reduces once at the end.
  - **`Field::square`** — new trait method. In characteristic 2, squaring is just bit-spread (`aᵢ → 2i`)
  with no multiplies. Benchmarks:
    - soft: ~7× faster than mul
    - x86: **bit-spread beats PCLMUL** for standalone Gf2_128 square — 11.5 ns → 6.0 ns. PCLMUL is retained
   (as `square_xmm`) only inside the inverse loop where state is already register-resident.
    - wasm: 4.2× faster than mul via parallel v128 bit-spread.
  - **XMM-resident inverse on x86** — the 127-iteration Fermat chain now keeps running state in `__m128i`
  throughout, collapsing 254 GPR↔XMM transfers to 2. Uses the 2-CLMUL squaring variant on the 127
  squarings.
  - **`rayon` feature** — `Field::inner_product` auto-parallelises when the input exceeds ~L2 cache size on
   a single core (byte-aware threshold: `64 KB × 2 / T::BYTE_SIZE`). Small inputs stay sequential to avoid
  thread-spawn overhead. No new public API — opting in via `--features rayon` is all that's needed.

  ## Measured

  On my machine:

  | op                              | before / sequential    | after         | speedup |
  | ------------------------------- | ---------------------- | ------------- | ------- |
  | `gf2_128/mul` (x86)             | schoolbook u128 loop   | 15.9 ns       | ~10×    |
  | `gf2_128/square` (x86)          | 15.9 ns (mul path)     | 6.0 ns        | 2.6×    |
  | `gf2_128/square` (soft)         | 44.6 ns (mul path)     | 6.0 ns        | 7.4×    |
  | `gf2_128/square` (wasm)         | 37.2 ns (mul path)     | 8.8 ns        | 4.2×    |
  | `gf2_128/inner_product` at 2^20 | 1.81 ms (single core)  | 281 µs (rayon) | 6.4×   |

  ## Test coverage

  - `REFERENCE_PRODUCTS` tables for both fields (20 vectors for Gf2_64, 12 for Gf2_128 — hand-derived and
  Intel white-paper vectors). The same truth table cross-validates x86, soft, and wasm backends
  bit-for-bit.
  - `test_field_square` — 1000 random `x.square() == x * x` checks per field.
  - `test_field_axioms_random` — 1000 random iterations of commutativity, associativity, distributivity,
  and inverse round-trip per field.
  - `test_field_inner_product` — lengths 0, 1, 17, 1024, 20_000 — the 20_000 case crosses the rayon
  parallel threshold, so it exercises the `par_chunks` branch when the feature is on.
  - `test_reduction_constant` — pins `x⁶⁴ ≡ 0x1b` and `x¹²⁸ ≡ 0x87` directly.
  - `#[should_panic]` tests for inner_product length mismatch.
  - `bmul.rs` direct unit tests for the soft primitives against hand-derived known-answer vectors.
  - Full test suite passes under `+pclmulqdq` (27 tests), `-pclmulqdq` soft (32), and
  `wasm32-wasip1-threads` via wasmtime (25, +2 ignored `should_panic`).

  ## Benches

  Consolidated into `benches/fields.rs`: mul / square / inverse / inner_product for both fields,
  inner_product sweeps `2^8 … 2^20` with `Throughput::Elements` so criterion reports elem/s. Replaces the
  previous `benches/inverse_gf2_128.rs`.


themighty1 (approved):
nice work, ack

PR #389, opened 2026-05-06
feat: add mpz-circuits-new crate with Context trait and SHA-256 circuit

sinui0:
A field-generic circuit-evaluation abstraction. The `Context` trait exposes `add` / `sub` / `mul` / `mul_const` / `constant` / `assert_const` / `assert_eq` over an associated `Wire` type, so the same circuit code drives plain-bit witness evaluation, ZK provers, and ZK verifiers without modification.

Includes:
- `Context` trait + `MaybeConst<Wire, Field>` return type for `mul_const` (skips communication when one operand is a known constant).
- `WitnessCtx` impl: plain-field arithmetic, AND-gate outputs appended to a witness tape.
- Hand-written SHA-256 compression circuit (`sha256::compress`) with the constant `AND_PER_BLOCK` and initial state `H0` exported.
- Criterion bench for SHA-256 over the witness context.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "crates/common",
     "crates/fields",
     "crates/circuits",
+    "crates/circuits-new",
     "crates/circuits-core",
     "crates/circuits-data",
     "crates/cointoss",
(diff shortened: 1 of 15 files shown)


sinui0:
@themighty1 you can rebase the polynomial proof PR on this and build against this abstraction. The `Context::Wire` associated type can be a polynomial

themighty1 (approved):

PR #398, opened 2026-05-19
Fix `yield_now` being noop on wasm

th4s:
This PR adds proper thread parking to the new executor using `std::thread::park`.


heeckhau:
@claude review

PR #403, opened 2026-05-26
refactor(common): unbundle executor, support global thread pool

sinui0:


sinui0:
@heeckhau as discussed earlier, it would be helpful if you could integration test these changes with TLSNotary to confirm that the max stream limit is no longer exceeded and there are no performance regressions :pray:

heeckhau:
@claude review

heeckhau (approved):
lgtm

Issue #448, opened 2026-08-01
Context::map exhausts a multiplexer's channel budget on large workloads


mpzFixApplier:
### What happens

`Context::map` gives each item its own child context, and each child context takes a channel from the multiplexer. `buffered(concurrency_limit)` (added in #403) bounds how many items are *in flight*, but each item still gets its own channel ID, so a map over N items opens N channels over its lifetime.

For a mux that tracks channels rather than in-flight work, that is still unbounded. A mux frees a channel when its stream is dropped, but the release is handled by the connection task and lags behind the rate at which the sliding window opens new ones so the live-channel count keeps climbing on a large enough workload and eventually hits the mux's cap.

### Where we hit it

Driving tlsn (alpha.15) MPC-TLS sessions. Its mux allows 512 concurrent streams (`crates/tlsn/src/session.rs`), and the item count for the circuit calls scales with the configured `max_sent_data`. Around a ~7 KB budget, roughly half of all sessions die:

```
context mux error
```

That message is the whole diagnostic, which is part of why this took a while to pin down: `ContextError`'s `Display` prints only its own text, and the source it hides is `maximum number of streams reached`. Adding `{source}` to that `Display` would make this class of failure self-explanatory happy to send it as a separate one-liner if you want it.

Measured on a loopback MPC-TLS run, sweeping `max_sent_data` with everything else equal:

| `max_sent_data` | with the `buffered` window |
|---|---|
| ~1 KB | no failures observed |
| 6144 - 7040 | intermittent |
| 7168 | 7/15 runs ok |
| 8192 and up | mostly failing |

Small maps are fine, which is why ordinary workloads don't see it — it needs enough items to outrun the mux's channel reclamation.

### Suggested fix

Bound the number of channels a map ever opens, not just how many items run at once: distribute items round-robin over at most `concurrency_limit` lanes, each lane owning one child context and processing its items sequentially. Channel usage becomes `min(items.len(), concurrency_limit)`, `concurrency_limit` keeps both its meaning and its default, and both parties still derive an identical layout from the item index alone.

With that change the same sweep gives 12/12 at 7168, 12/12 at 8192 and 10/10 at 16384.

I have this implemented and tested against `dev` (`cargo test --workspace` green, existing `test_map_respects_concurrency_limit` still holds since lanes bound concurrency too) and will open a PR referencing this issue. Happy to take it in a different direction if you would rather solve it inside the mux, or by making the channel-id scheme reuse ids across the window.

PR #449, opened 2026-08-01
fix(common): bound the channels Context::map opens, not just its concurrency

mpzFixApplier:
Fixes the channel exhaustion described in #448  

### The problem

`Context::map` gives every item its own child context, and therefore its own multiplexer channel. `buffered(concurrency_limit)` bounds how many items run at once, but the channel IDs are still per item, so a map over N items opens N channels over its lifetime.

That is enough to exhaust a mux which tracks channels rather than in-flight work. A mux frees a channel when its stream is dropped, but that release is processed by the connection task and lags behind the rate at which the sliding window opens new ones, so the live-channel count keeps climbing. Against tlsn's mux (512 streams) an MPC-TLS session with a ~7 KB `max_sent_data` budget fails this way in roughly half of all runs, surfacing as `context mux error`  whose hidden source is `maximum number of streams reached`.

### The change

Distribute items round-robin over at most `concurrency_limit` *lanes*, each lane owning one child context and processing its items sequentially.

- Channels ever opened: `min(items.len(), concurrency_limit)`, independent of workload size.
- That is still the concurrency bound too, so `concurrency_limit` keeps its meaning — and its default of 32 is unchanged here.
- Lane assignment is `index % lanes`, derived from the item index alone, so both parties agree on the channel layout and per-channel message order with no extra coordination.
- Results are still returned in input order.

### Measurements

tlsn alpha.15 driving loopback MPC-TLS at a given `max_sent_data`, everything else equal:

| | `buffered` window | round-robin lanes |
|---|---|---|
| MAX_SENT=7168 | 7/15 ok | 12/12 ok |
| MAX_SENT=8192 | mostly failing | 12/12 ok |
| MAX_SENT=16384 | mostly failing | 10/10 ok |

### Tests

`cargo test --workspace` passes. The existing `test_map_respects_concurrency_limit` still holds unchanged (lanes bound concurrency as well), and a two-party recording test is added that maps far more items than the limit and pins the opened-channel count to the lane count while checking both sides stay in lockstep and results come back in input order.

### Changelog

Added under `## [Unreleased]` -> `### Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,3 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Fixed
+- `Context::map` no longer opens one multiplexer channel per item. Items are distributed
+  round-robin over at most `concurrency_limit` lanes, so the number of channels a `map` opens
+  is bounded by the limit rather than by the size of the workload.
--- a/crates/common/src/context.rs
+++ b/crates/common/src/context.rs
@@ -8,7 +8,6 @@ use std::sync::Arc;
 use futures::{
     AsyncRead, AsyncWrite,
     future::{self, BoxFuture, Either},
-    stream::{self, StreamExt, TryStreamExt},
 };
 
 #[cfg(any(test, feature = "test-utils"))]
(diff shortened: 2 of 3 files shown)


themighty1:
Thanks for the PR it was merged in https://github.com/tlsnotary/mpz-1/
which is a temporary fork of this repo which TLSNotary will use until the ethereum org grants me admin permissions to this repo or it will be transfered under the tlsnotary org. Assuming it may take a while, hence the need for a fork.

themighty1:
Checked in'tlsn' harness for regressions, good to go.

